PR #12, opened 2021-11-30
Fixes cropped carousel items in Safari

danieldietrich:
This fixes a deviation of the carousel rendering between Chrome and Safari.

# ⬅️ Chrome | Safari ➡️

## Before
Before it looked on large screens like this (left: Chrome, right: Safari):

before-lg

## After

After applying this fix, it looks correct:

after-lg

This fix does not change rendering on small screens. Before and after look like this:

before-after-sm

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -61,5 +61,5 @@
 }
 
 .feature-item-content {
-    @apply sm:absolute h-full w-full box-border rounded-xl border-8 border-emeraldLangium dark:border-emeraldLangium sm:px-6 px-2 pb-6 mx-2 bg-emeraldLangium dark:bg-gray-900 shadow-lg;
+    @apply h-full w-full box-border rounded-xl border-8 border-emeraldLangium dark:border-emeraldLangium sm:px-6 px-2 pb-6 mx-2 sm:mx-0 bg-emeraldLangium dark:bg-gray-900 shadow-lg;
 }
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1454,7 +1454,8 @@ video {
 
 @media (min-width: 640px) {
   .feature-item-content {
-    position: absolute;
+    margin-left: 0px;
+    margin-right: 0px;
   }
 
   .feature-item-content {


public/css/style.css
```diff
@@ -802,8 +802,8 @@ video {
   height: 9rem;
 }
 
-.h-64 {
-  height: 16rem;
+.h-60 {
+  height: 15rem;
```
danieldietrich, line 806: This shouldn't be affected by this PR. I guess that this file wasn't committed in a prev PR...

public/css/style.css
```diff
@@ -818,10 +818,6 @@ video {
   height: 4rem;
 }
 
-.h-60 {
-  height: 15rem;
-}
-
```
danieldietrich, line 824: dito

src/style.css
```diff
@@ -61,5 +61,5 @@
 }
 
 .feature-item-content {
-    @apply sm:absolute h-full w-full box-border rounded-xl border-8 border-emeraldLangium dark:border-emeraldLangium sm:px-6 px-2 pb-6 mx-2 bg-emeraldLangium dark:bg-gray-900 shadow-lg;
+    @apply h-full w-full box-border rounded-xl border-8 border-emeraldLangium dark:border-emeraldLangium sm:px-6 px-2 pb-6 mx-2 sm:mx-0 bg-emeraldLangium dark:bg-gray-900 shadow-lg;
```
danieldietrich, line 64: Using `position:absolute` looks like a hack for removing the margin in the presence of `mx-2`. Removing that and properly setting horizontal margin to zero fixed the issue.

public/css/style.css
msujew: I think you need to rebase on the current `main`, the previous one was missing all of the hugo changes.

danieldietrich: thanks @msujew that worked _(I wasn't aware that main wasn't the source of truth)_

jbicker (approved):
Works perfectly! Thank you!

danieldietrich:
Thx Jan

PR #21, opened 2022-01-14
Grammar documentation

gfontorbe:
Here is a first draft to the grammar language section of the Langium documentation.

Open for any suggestion about the content, structure, etc.

Most likely some parts need to be reworked to be in line with v0.3

### DIFF
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -240,7 +240,7 @@ but the following:
 person Bob
 Hello Sara !
 ```
-will result in an error message since the cross reference resolution will fail because a `Person` object with the ID of 'Sara' has not been instantiated even though 'Sara' is a valid `ID`.
+will result in an error message since the cross reference resolution will fail because a `Person` object with the name 'Sara' has not been defined, even though 'Sara' is a valid `ID`.
 
 #### Unordered Groups
 [NOTE] Unordered group are currently not supported but seems like it's being worked on. I think this is still valuable info that could be added to the documentation at a later stage if needed.
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -357,7 +357,7 @@ The *entry rule* is a parser rule that defines the starting point of the parsing
  ```
 In this example, the *entry rule* `Model` defines a group of alternatives. The parser will go through the input document and try to parse a `Person` or a `Greeting` object and add it to the array `persons` or `greetings`, respectively. Thanks to the grouping and the cardinality of zero or many (`*`), the parser will go through the document until all inputs have been consumed.
 
-Not all parser rules need to be registered in the parser rule.
+Not all parser rules need to be registered in the entry parser rule.
 ```
 entry Model:
     (persons+=Person | greetings+=Greeting)*;
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -68,19 +68,21 @@ A cardinality defines the number of elements in a given set. Four different card
 
 [QUESTION]Should we add examples of cardinalities here?
 #### Character Range
-The operator `..` is used to declare a character range. It is equivalent to the operator `-`in regular expression.
+The operator `..` is used to declare a character range. It is equivalent to the operator `-`in regular expression. It matches any character in between the left character and the right character (inclusive on both ends) depending on their UTF-16 code.
+
+[QUESTION] I could find where in the langium code base we parse this range. I assumed it would be based on String.prototype.charCodeAt() or something similar. Please correct if I'm wrong.
 
 ```
 terminal INT returns number: ('0'..'9')+;
 ```
 
-is equivalent to
+is equivalent to the regular expression:
 
 ```
 terminal INT returns number: /[0-9]+/;
 ```
 
-Here, INT is matched to one or more characters (by using the operand `+`) between '0' and '9' (inclusive).
+Here, INT is matched to one or more characters (by using the operand `+`, which defines a cardinality of 'one or many') between `0` and `9` (inclusive on both ends).
 
 #### Wildcard
 The operator `.` is used to match any character and is similar in regular expression.
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -60,7 +60,7 @@ A cardinality defines the number of elements in a given set. Four different card
 4. one or many (operator `+`)
 
 #### Character Range
-The operator `..` is used to declare a character range. It is equivalent to the operator `-` in regular expression. It matches any character in between the left character and the right character (inclusive on both ends).
+The operator `..` is used to declare a character range. It is equivalent to the operator `-` within a character class in a regular expression. It matches any character in between the left character and the right character (inclusive on both ends).
 ```
 terminal INT returns number: ('0'..'9')+;
 ```
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -190,13 +190,17 @@ There are three different ways to assign an expression (right side) to a propert
     Paragraph:
         'paragraph' (sentences+=STRING)+;
     ```
-    [QUESTION] IS THIS DECLARATION VALID? OR SHOULD sentences BE ASSIGNED AN OTHER PARSER RULE INSTEAD OF A TERMINAL RULE?
-3. `?=` is used to assign a **value to a property of type boolean**. The value of the property of type `boolean` is set to `true` if the right part of the assignment is consume by the parser
+    Here, the property `sentences` will accept one or many expressions matching the terminal rule `STRING`. The expressions belong to a single group and must not be interrupted by other expressions. The parsing of a document containing:
+    ```
+    paragraph "The expression group " 3 "was interrupted"
+    ```
+    will throw an error since the `STRING` expressions are not continuous. It is however possible to interrupt and resume a sequence of expressions by using [hidden terminal symbols](#hidden-terminal-symbols).
+3. `?=` is used to assign a **value to a property of type boolean**. The value of the property of type `boolean` is set to `true` if the right part of the assignment is consumed by the parser
     ```
     Employee:
         'employee' name=ID (remote?='remote')?
     ```
-    Here the value of the property `remote` will be set to true if the keyword `remote` appears in the declaration.
+    Here the value of the property `remote` will be set to true if the keyword `remote` appears in the document.
 #### Cross-reference
 [TODO] CHECK IN MORE DETAILS, I DON'T UNDERSTAND THE SYNTAX AND HOW IT'S IMPLEMENTED
 #### Unordered Groups
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -121,8 +121,9 @@ terminal BETWEEN_HASHES: /#[^#]*#/;
 #### Rule Calls
 A terminal rule can implement other terminal rules. 
 ```
-terminal DOUBLE: INT '.' INT;
+terminal DOUBLE returns number: INT '.' INT;
 ```
+Note: it is easy to create conflict between terminal rules by using *terminal rule calls*. See [Data Type Rules](#data-type-rules) for further details.
 #### Alternatives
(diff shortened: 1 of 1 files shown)


content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+```
+terminal STRING: /"[^"]*"|'[^']*'/;
+```
+
+#### Groups
+Tokens can be put in sequence specifying the order they have to appear
+```
+terminal FLIGHT_NUMBER: ('A'..'Z')('A'..'Z')('0'..'9')('0'..'9')('0'..'9')('0'...'9')?;
+```
+In this example, the token `FLIGHT_NUMBER` **must** start with two capital letters and followed by three of four digits.
```
msujew, line 119: Minor:
```suggestion
In this example, the token `FLIGHT_NUMBER` **must** start with two capital letters and followed by three or four digits.
```

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+
+### Terminal Fragments
+Fragments allow for sub-definition of terminal rules to be extracted. They are not consumed by the lexer and have to be consumed by other terminal rules.
```
msujew, line 122: Suggestion: Add here, that normal terminals can also be used as part of other terminals, not only fragments.

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+```
+
+## Parser Rules
+While [terminal rules](#terminal-rules) indicate to the lexer what sequence of characters are valid tokens, *parser rules* indicate to the parser what sequence of tokens are valid. Parser rules lay the structure of objects to be created by the parser and result in the creation of the *abstract syntax tree* which represents the syntactic structure of your language. In Langium, parser rules are also responsible for defining the type of objects to be parsed.
+
+### Extended Backus-Naur Form Expressions
+Parser rules have access to a more restricted set of EBNF expressions compared to terminal rules. Parser and terminal rules share the following expressions:
+- Groups
+- Alternatives
+- Keywords
```
msujew, line 145: Terminal rules are also able to use keywords. `terminal X: 'Y';` is a valid terminal.

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+    paragraph "expression one" /* comment */ "expression two"
+    ```
+    The above example will be successfully parsed.
+
+3. `?=` is used to assign a **value to a property of type boolean**. The value of the property of type `boolean` is set to `true` if the right part of the assignment is consumed by the parser.
+    ```
+    Employee:
+        'employee' name=ID (remote?='remote')?
+    ```
+    Here the value of the property `remote` will be set to true if the keyword `remote` appears in the document.
```
msujew, line 216: Minor: This is a bit misleading.

```suggestion
    Here the value of the property `remote` will be set to `true` if the keyword `remote` was successfully parsed as a part of the rule call.
```

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+
+#### Cross-reference
+With Langium, you can declare *cross-references* directly in the grammar. *Cross-reference* allows to reference an object of a given type. The syntax is:
```
msujew, line 219: Minor:

```suggestion
With Langium, you can declare *cross-references* directly in the grammar. A *cross-reference* allows to reference an object of a given type. The syntax is:
```

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+```
+property=[Type:TOKEN]
+```
+The `property` will be a reference to an object of type `Type` identified by the token `TOKEN`. If the `TOKEN` is omitted, the parser assumes that it is matching the terminal rule `ID`.
```
msujew, line 223: Since we use a lot of functionality in this documentation which is only available in version 0.3, we should use the 0.3 behavior here as well:

```suggestion
The `property` will be a reference to an object of type `Type` identified by the token `TOKEN`. If the `TOKEN` is omitted, the parser will use the terminal or data type rule associated with the `name` assignment of the `Type` rule.
```

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+    'person' name=ID;
+Greeting:
+    'Hello' person=[Person:ID] '!';
+```
+The `Person` in square brackets does not refer to the parser rule `Person` but instead refers to an object of type `Person`. It will successfully parse a document like:
+```
+person Bob
+Hello Bob !
+```
+but will fail to parse:
```
msujew, line 235: Suggestion: The _parse_ will not fail, since this isn't a parsing error. Instead, the cross reference resolution will fail, resulting an error message.

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+Person:
+    'person' name=ID & age=INT
+```
+will now allow `name` and `age` to be declared in any order.
+```
+person 25 Bob
+```
+will then successfully create an object of type `Person`.
+
+Cardinality (?,*,+ operators) also applies to unordered group. Please note that assignments with a cardinality of `+` or `*` have to appear continuously and cannot be interrupted by an other assignment and resumed later.
```
msujew, line 271: Minor: Escape usage of `*` using `\*`. This breaks markdown formatting.

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+#### Unassigned Rule Calls
+Parser rules do not necessarily need to return an object, they can also refer to other parser rules which in turn will be responsible for returning the object.
+For example, in the [Arithmetics example](https://github.com/langium/langium/blob/main/examples/arithmetics/src/language-server/arithmetics.langium):
+```
+AbstractDefinition:
+	Definition | DeclaredParameter;
+```
+The parser rule `AbstractDefinition` will not create an object of type AbstractDefinition. Instead, it calls either the `Definition` or `DeclaredParameter` parser rule which will be responsible for creating an object of a given type (or call other parser rules if they are unassigned rule calls themselves). 
+
+#### Assigned Actions
```
msujew, line 297:
```suggestion
#### Assignments
```

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+Expression:
+    '(' Addition ')' | value=INT;
+
+Addition:
+    Expression ('+' Expression)*;
+```
+
+[NOTE] I left out the part regarding `tree rewrite actions` as I think it could be part of a more advanced documentation part on the AST/CST. I can add it if we think it belongs here.
+
+### Syntactic Predicates
```
msujew, line 316: Syntactic predicates were only needed in Xtext due to limitations of the underlying parser generator (Antlr3). We don't need syntactic predicates in Langium. We support them, but only for backwards compatibility.

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+
+## The Entry Rule
+The *entry rule* is a parser rule that defines the starting point of the parsing step. The *entry rule* starts with the keyword `entry` and matches other parser rules.
+ ```
+ entry Model:
+    (persons+=Person | greetings+=Greeting)*;
+ ```
+In this example, the *entry rule* `Model` defines a group of alternatives. The parser will go through the input document and try to parse a `Person` or a `Greeting` object and add it to the array `persons` or `greetings`, respectively. Thanks to the grouping and the cardinality of zero or many (`*`), the parser will go through the document until all inputs have been consumed.
+
+Not all parser rules need to be registered in the parser rule.
```
msujew, line 357: Minor:

```suggestion
Not all parser rules need to be registered in the entry parser rule.
```

msujew (commented):
Great, that looks really good as a start 👍

content/docs/grammar-language.md
gfontorbe: Isn't that what is documented in the "rule calls" section (101-106) ? Or am I missing something?

content/docs/grammar-language.md
```diff
@@ -0,0 +1,371 @@
...
+Equivalent in regular expression:
+```
+terminal HASHTAG: /#.+/;
+```
+#### Until Token
+The operator `->` indicates that all characters should be consumed from the left token *until* the right token occurs. For example, the terminal rule for multi-line comment can be implemented as:
+```
+terminal ML_COMMENT: '/*' -> '*/';
+```
+There is no direct equivalent in regular expression for the until token.
```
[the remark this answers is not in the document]
gfontorbe: Added equivalence in commit [5487c8f](https://github.com/langium/langium-website/pull/21#:~:text=until%20token%20equivalence-,5487c8f,-Merge%20state)

content/docs/grammar-language.md
gfontorbe: https://github.com/langium/langium-website/pull/21#:~:text=cross%20reference%20resolution-,847ceb2,-Merge%20state

content/docs/grammar-language.md
gfontorbe: It looks fine on the Netlify preview

content/docs/grammar-language.md
gfontorbe: There is already a section "Assignments" (line ~190) before where it is described how to assign data to a property/feature. 
Any idea on renaming that wouldn't be confusing with the previous section?

content/docs/grammar-language.md
gfontorbe: Should we then mention that it is supported for backwards compatibility but not recommend it for new grammars? Can we then just link to the Xtext documentation?

msujew: The grammar is backwards compatible in a way that enables syntactic predicates from a grammar parser perspective , but it does not do anything in any way. So I think we shouldn't mention it at all.

gfontorbe: Removed

content/docs/grammar-language.md
```diff
@@ -0,0 +1,351 @@
...
+    'keywordTwo' name=ID;
+```
+In the above example, we rely on a *rule call* to specify the return type. With more complex rules, the readability will be highly impacted. *Actions* allow to improve the readability of the grammar by explicitly defining the return type. *Actions* are declared inside of curly brackets `{}`:
+```
+RuleName returns TypeOne:
+    'keywordOne' name=ID | 'keywordTwo' {TypeTwo} name=ID;
+```
+
+#### Unassigned Rule Calls
+Parser rules do not necessarily need to return an object, they can also refer to other parser rules which in turn will be responsible for returning the object.
```
spoenemann, line 292: Proposal: move this subsection before _Simple Actions_

```suggestion
Parser rules do not necessarily need to create an object, they can also refer to other parser rules which in turn will be responsible for returning the object.
```

content/docs/grammar-language.md
```diff
@@ -0,0 +1,351 @@
...
+    '(' Addition ')' | value=INT;
+
+Addition:
+    Expression ('+' Expression)*;
+```
+
+[NOTE] I left out the part regarding `tree rewrite actions` as I think it could be part of a more advanced documentation part on the AST/CST. I can add it if we think it belongs here.
+
+## Data Type Rules
+Data type rules are similar to terminal rules as they match a sequence of characters. However, they are parser rules and therefore are context-dependent, and are allowed to use hidden terminal rules. Contrary to terminal rules, they cannot use *regular expressions* to match a stream of character and have to compose with terminal rules.
```
spoenemann, line 320:
```suggestion
Data type rules are similar to terminal rules as they match a sequence of characters. However, they are parser rules and therefore are context-dependent, and are allowed to use hidden terminal rules. Contrary to terminal rules, they cannot use *regular expressions* to match a stream of character and have to be composed of terminal rules and keywords.
```

spoenemann (commented):
Good work, I like the writing style! It is mostly very clear.

content/docs/grammar-language.md
spoenemann: Right, but we don't call those keywords. They are merely character sequences (like groups): `'xyz'` is the same as `'x' 'y' 'z'`.

spoenemann (approved):
Thanks a lot! I'll read it through once again after merging and make further changes right away if I find more things to improve.

PR #22, opened 2022-01-27
Added draft dependency injection

gfontorbe:
First draft for the documentation about configuration of services via dependency injection.

It has two main parts:
1. Adding new services
2. Extending/Overriding existing services

The last part is a brief description of all default services and their dependencies.

### DIFF
--- a/content/docs/configuration-services.md
+++ b/content/docs/configuration-services.md
@@ -118,7 +118,7 @@ MyService: () => new MyNewServiceClass()
 ```
 Now, every call to `MyService` will create an instance of `MyNewServiceClass` instead of `DefaultServiceClass`.
 
-In real scenarios, it is not recommend to completely replace a service. Instead, it is possible to extend a class and override only specific parts of it. Let's take a look at the `ArithmeticsModule` from the [arithmetics example](https://github.com/langium/langium/tree/main/examples/arithmetics).
+In most cases, it is not necessary to completely replace a service. Instead, it is possible to extend a class and override only specific parts of it. Let's take a look at the `ArithmeticsModule` from the [arithmetics example](https://github.com/langium/langium/tree/main/examples/arithmetics).
 
 ```Typescript
 export const ArithmeticsModule: Module<ArithmeticsServices, PartialLangiumServices & ArithmeticsAddedServices> = {
--- a/content/docs/configuration-services.md
+++ b/content/docs/configuration-services.md
@@ -16,12 +16,9 @@ The *shared modules* contain services shared across all languages:
 
 ## Language Specific Modules
 The *language specific modules* contain all services which are specific to one language:
-* Services to build the `parser` for a specific language.
-* Services to handle *LSP requests*.
-* Services to retrieve information from a given AST.
-* Services to manage references during linking and inside of documents.
-* Serialization and deserialization of objects to JSON.
-* Services to handle document and syntax validation.
+* Services for [LSP features](#language-server-protocol)
+* Services involved in the document lifecycle (future documentation)
+* Utility services
 
 ## Customization
 If you used the [Yeoman generator](https://www.npmjs.com/package/generator-langium), the entry point to services customization is found in the `src/language-server/...-module.ts` file, where '...' is the name of your language. There, you can register new services, or override the default implementation of services. Langium implements the *Inversion of Control* principle via *Dependency Injection* pattern, which promotes loosely-coupled architectures, maintainability, and extensibility.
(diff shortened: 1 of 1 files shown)


content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+
+### DefaultModule
+The `DefaultModule` contains services specific to one unique language. These are the basic services such as configuring the parser or the LSP.
+#### GrammarConfig
+#### GrammarConfig
+The `GrammarConfig` service is responsible for extracting specific information from the `Grammar`. By default, the `GrammarConfig` service extracts rules for multiline comments.
+
+*Dependencies:* [Grammar](#grammar)
+#### LangiumParser
+The `LangiumParser` service is responsible for creating the parser. It extracts information from the grammar and creates the parser using Chevrotain.
```
sailingKieler, line 236:
```suggestion
The `LangiumParser` is set up by inspecting the rules and definitions in the grammar. The implementation is based on Chevrotain.
```

Maybe link Chevrotain?

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+
+*Dependencies:* [Grammar](#grammar) | [TokenBuilder](#tokenbuilder) | [LanguageMetaData](#languagemetadata) | [Linker](#linker) | [ValueConverter](#valueconverter)
+#### ValueConverter
+The `ValueConverter` service is responsible for converting an input into its correct primitive type. The parser parses the input file as a collection of strings which need to be converted to their respective type.
+#### TokenBuilder
+The `TokenBuilder` service is responsible for creating an array of `TokenType` from the grammar file.
```
sailingKieler, line 242: Is this worth mentioning?

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+#### CompletionProvider
+The `CompletionProvider` service is responsible for handling a *Completion Request* at a given cursor position. When a *Completion Request* is submitted by the client to the server, the `CompletionProvider` will create a `CompletionList` of all possible `CompletionItem` to be presented in the editor. The `CompletionProvider` service computes a new `CompletionList` after every typing.
+
+*Dependencies:* [ScopeProvider](#scopeprovider) | [RuleInterpreter](#ruleinterpreter) | [Grammar](#grammar)
+#### RuleInterpreter
+The `RuleInterpreter` service is used by the `CompletionProvider` service to identify any `AbstractElement` that could be present at a given cursor position. The parser uses the best-fitting grammar rule for a given text input. However, if there are multiple rules that could be applied, only one of them will be successfully parsed. The `RuleInterpreter` service solves this issue by returning **all** possible features that could be applied at a cursor position.
```
sailingKieler, line 248: I would drop this here as well, as it's to low-level

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+#### LangiumParser
+The `LangiumParser` service is responsible for creating the parser. It extracts information from the grammar and creates the parser using Chevrotain.
+
+*Dependencies:* [Grammar](#grammar) | [TokenBuilder](#tokenbuilder) | [LanguageMetaData](#languagemetadata) | [Linker](#linker) | [ValueConverter](#valueconverter)
+#### ValueConverter
+The `ValueConverter` service is responsible for converting an input into its correct primitive type. The parser parses the input file as a collection of strings which need to be converted to their respective type.
+#### TokenBuilder
+The `TokenBuilder` service is responsible for creating an array of `TokenType` from the grammar file.
+#### CompletionProvider
+The `CompletionProvider` service is responsible for handling a *Completion Request* at a given cursor position. When a *Completion Request* is submitted by the client to the server, the `CompletionProvider` will create a `CompletionList` of all possible `CompletionItem` to be presented in the editor. The `CompletionProvider` service computes a new `CompletionList` after every typing.
```
sailingKieler, line 244:
```suggestion
The `CompletionProvider` service is responsible for handling *LSP Completion Requests* at a given cursor position. When a *Completion Request* is submitted by the client to the server, the `CompletionProvider` will create a `CompletionList` of all possible `CompletionItem` to be presented in the editor. The `CompletionProvider` service computes a new `CompletionList` after every typing.
```

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+
+*Dependencies:* [ScopeProvider](#scopeprovider) | [RuleInterpreter](#ruleinterpreter) | [Grammar](#grammar)
+#### RuleInterpreter
+The `RuleInterpreter` service is used by the `CompletionProvider` service to identify any `AbstractElement` that could be present at a given cursor position. The parser uses the best-fitting grammar rule for a given text input. However, if there are multiple rules that could be applied, only one of them will be successfully parsed. The `RuleInterpreter` service solves this issue by returning **all** possible features that could be applied at a cursor position.
+#### DocumentSymbolProvider
+The `DocumentSymbolProvider` service is responsible for handling a *Document Symbols Request*. The `DocumentSymbolProvider` is used to return a hierarchy of all symbols found in a document as an array of `DocumentSymbol`.
```
sailingKieler, line 250:
```suggestion
The `DocumentSymbolProvider` service is responsible for handling *LSP Document Symbols Requests*. The `DocumentSymbolProvider` is used to return a hierarchy of all symbols found in a document as an array of `DocumentSymbol`. 
```

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+#### DocumentSymbolProvider
+The `DocumentSymbolProvider` service is responsible for handling a *Document Symbols Request*. The `DocumentSymbolProvider` is used to return a hierarchy of all symbols found in a document as an array of `DocumentSymbol`. 
+
+*Dependencies:* [NameProvider](#nameprovider)
+#### HoverProvider
+The `HoverProvider` service is responsible for handling a *Hover Request* at a given text document position. By default, Langium implements the possibility to generate tooltips with the content of a multiline comment while hovering a symbol.
+
+*Dependencies:* [References](#references)
+#### FoldingRangeProvider
+The `FoldingRangeProvider` service is responsible for handling a *Folding Range Request*. This service identifies all the blocks that can be folded in a document. By default, Langium adds the possibility to fold multiline comments.
```
sailingKieler, line 258: same here

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+#### ReferenceDescriptionProvider
+The `ReferenceDescriptionProvider` is responsible for creating a description for a given reference. By default the description includes the `sourceUri`, `sourcePath`, `targetUri`, `targetPath`, `segment`, and if the reference is `local`.
+
+*Dependencies:* [Linker](#linker) | [AstNodeLocator](#astnodelocator)
+#### Linker
+The `Linker` service is responsible for resolving cross-references in the AST.
+
+*Dependencies:* [AstReflection](#astreflection) | [LangiumDocuments](#langiumdocuments) | [ScopeProvider](#scopeprovider) | [AstNodeLocator](#astnodelocator)
+#### NameProvider
+The `NameProvider` service is responsible for getting the name of a given `AstNode`. If the `AstNode` does not have a name, the `NameProvider` returns `undefined`.
```
sailingKieler, line 288: Maybe it's worth noting that it is a direct accomplice of the `AstNodeDescriptionProvider` (although that can be seen on the dependencies.

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+#### ScopeProvider
+
+*Dependencies:* [AstReflection](#astreflection) | [IndexManager](#indexmanager)
+#### ScopeComputation
+
+*Dependencies:* [NameProvider](#nameprovider) | [AstNodeDescriptionProvider](#astnodedescriptionprovider)
+#### References
+The `References` service is responsible for finding all references to the target node.
```
sailingKieler, line 296:
```suggestion
The `References` service is responsible for finding all references to some target node, as e.g. offered in VS Code's text editors.
```

Does that appendix make sense to you?

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+
+*Dependencies:* [NameProvider](#nameprovider) | [IndexManager](#indexmanager) | [AstNodeLocator](#astnodelocator)
+#### JsonSerializer
+The `JsonSerializer` service is responsible for serializing and deserializing JSON.
```
sailingKieler, line 300: Is that any easy to see use case for that?

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+The `DocumentValidator` service is responsible for validating an entire document.
+
+*Dependencies:* [ValidationRegistry](#validationregistry)
+#### ValidationRegistry
+The `ValidationRegistry` service manages a set of `ValidationCheck` that are defined inside of a `Validator` service.
+
+*Dependencies:* [AstReflection](#astreflection)
+
+### LanguageGeneratedModule
+The `LanguageGeneratedModule` contains services specific to the grammar of a given language. They are automatically generated by Langium and should not be modified.
```
sailingKieler, line 313:
```suggestion
The `LanguageGeneratedModule` contributes language specific services. It automatically re-generated by Langium and should not be modified manually.
```

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+#### Grammar
+The `Grammar` service holds a collection of terminal and parser rules defined by the language grammar.
```
sailingKieler, line 315:
```suggestion
The `Grammar` service provides information on the lexer and parser rules defined in the language grammar.
```

sailingKieler (commented):
Hi @gfontorbe,

first of all I very much appreciate your effort on the documentation. 👏
I have quite a bunch of suggestions, and I tried to make them 'constructive' (not just blaming ;-) ), please don't get them wrong. (I can be passionate while correcting and improving given stuff, whereas starting with a blank document is quite a challenge for me.)

If you have question, just ping me. Content-wise @msujew and @spoenemann will be more competent I guess.

sailingKieler:
One more thing:

Personally, I would switch the
```
### Adding New Services
### Overriding and Extending Services
```
sections, I guess that customizing the tooling by adjusting and overriding the existing services is more relevant or comes first when building language tooling, while introducing full-custom services comes second or even third.

@msujew and @spoenemann what do you think?

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+
+Let's say that you want to fully replace a service named `MyService` which implements `MyServiceInterface` interface and instantiate an instance of the class `DefaultServiceClass`.
+
+You would find the module declaration with the following line inside:
+```Typescript
+MyService: () => new DefaultServiceClass()
+```
+And `DefaultServiceClass` declared as:
+```Typescript
+export class DefaultServiceClass implements MyServiceInterface {
```
spoenemann, line 105: Usually the interface has the same name as the service itself, so without the `Interface` suffix.

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+    /* service logic */
+}
+```
+You then need to change how the module instantiate your service:
+```Typescript
+MyService: () => new MyNewServiceClass()
+```
+Now, every call to `MyService` will create an instance of `MyNewServiceClass` instead of `DefaultServiceClass`.
+
+In real scenarios, it is not recommend to completely replace a service. Instead, it is possible to extend a class and override only specific parts of it. Let's take a look at the `ArithmeticsModule` from the [arithmetics example](https://github.com/langium/langium/tree/main/examples/arithmetics).
```
spoenemann, line 121:
```suggestion
In most cases, it is not necessary to completely replace a service. Instead, it is possible to extend a class and override only specific parts of it. Let's take a look at the `ArithmeticsModule` from the [arithmetics example](https://github.com/langium/langium/tree/main/examples/arithmetics).
```

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+        const validator = services.validation.ArithmeticsValidator;
+        const checks: LangiumGrammarChecks = {
+            BinaryExpression: validator.checkDivByZero,
+            Definition: validator.checkNormalisable
+        };
+        this.register(checks, validator);
+    }
+}
+```
+Inside of the `ArithmeticsValidationRegistry`, we call our `ArithmeticsValidator` via the `ServiceRegistry` with `const validator = services.validation.ArithmeticsValidator` which will create a new instance of `ArithmeticsValidator`. Then, we declare the `checks` to be registered and register them inside of the registry via the function `register` which is declared in the parent class. The `ArithmeticsValidationRegistry` only adds functionality to the `ValidationRegistry` but does not override any functionality from it.
```
spoenemann, line 163: This part is quite specific to validation. Maybe extract that to a different section to keep "Overriding and Extending Services" more concise?

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+* Implementations of the `AstReflection` service (one implementation per language) to access the structure of the AST.
+
+## Language Specific Modules
+The *language specific modules* contain all services which are specific to one language:
+* Services to build the `parser` for a specific language.
+* Services to handle `LSP requests`.
+* Services to retrieve information from a given AST.
+* Services to manage references during linking and inside of documents.
+* Serialization and deserialization of objects to JSON.
+* Services to handle document and syntax validation.
```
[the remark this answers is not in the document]
spoenemann: Sounds good, but then the list would become quite long. There is already something like that at the end (Default Services).

content/docs/configuration-services.md
spoenemann: We use it in Langium to serialize and parse the grammar. This might also be useful to export ASTs to other tools or to the IDE client.

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+## Customization
+The entry point to services customization is found in the `src/language-server/...-module.ts` file, where '...' is the name of your language. There, you can register new services or override the default implementation of services. Langium implements the *Inversion of Control* pattern via *Dependency Injection* which promotes loosely-coupled architecture, maintainability and extensibility.
+
+For the following sections, we will use the [arithmetics example](https://github.com/langium/langium/tree/main/examples/arithmetics) to describe the procedure for adding or replacing services. Note that all notation containing *Arithmetics* should be understood as being specific to the language named *Arithmetics* and should be replaced with your own language name.
+
+### Adding New Services
+To add services which are not implemented by default by Langium, we first need to edit the type `ArithmeticsAddedService`.
+A language with no added services would have the type `ArithmeticsAddedService` declared as:
+```Typescript
+export type ArithmeticsAddedServices = {}
```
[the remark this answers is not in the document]
Pluralia: It is such way for all examples: [domainmodel](https://github.com/langium/langium/blob/5fc873db5aa5ae7897f6f3aa73ffa697ee14771b/examples/domainmodel/src/language-server/domain-model-module.ts#L14-L18), [statemachine](https://github.com/langium/langium/blob/5fc873db5aa5ae7897f6f3aa73ffa697ee14771b/examples/statemachine/src/language-server/statemachine-module.ts#L17-L21).
How I understand, the documentation has correct things and a validator has to be removed from `...AddedServices` everywhere.

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+The *language specific modules* contain all services which are specific to one language:
+* Services to build the `parser` for a specific language.
+* Services to handle `LSP requests`.
+* Services to retrieve information from a given AST.
+* Services to manage references during linking and inside of documents.
+* Serialization and deserialization of objects to JSON.
+* Services to handle document and syntax validation.
+
+## Customization
+The entry point to services customization is found in the `src/language-server/...-module.ts` file, where '...' is the name of your language. There, you can register new services, or override the default implementation of services. Langium implements the *Inversion of Control* pattern via *Dependency Injection*, which promotes loosely-coupled architectures, maintainability, and extensibility.
```
Pluralia, line 27: How I understand IoC is a principle and DI is a pattern, so: "Langium implements the Inversion of Control principle via the Dependency Injection pattern".

content/docs/configuration-services.md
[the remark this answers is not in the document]
gfontorbe: I would go for not adding that here if we keep the last part of the documentation in place.

content/docs/configuration-services.md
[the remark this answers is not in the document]
gfontorbe: The generator adds a LanguageValidator and overrides the ValidationRegistry by default. I will try to merge the 2 parts so it is less confusing.

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+```
+In the `ArithmeticsModule`, we map a property with the name of our service (here `ArithmeticsValidator`) to a concrete implementation of the service. This means that every time we need to call the service named "ArithmeticsValidator", a new instance of the class `ArithmeticsValidator` will be created. 
+
+The `ArithmeticsValidator` service does not depend on other services, and no argument is passed during the instantiation of the class. If you implement a service which depends on other service, you need to pass all services from the `ServiceRegistry` as an argument to the arrow function such as:
+```Typescript
+export const ArithmeticsModule: Module<ArithmeticsServices, PartialLangiumServices & ArithmeticsAddedServices> = {
+    ServiceWithDependencies = (services) => new ServiceClass(services);
+}
+```
+The services which `ServiceClass` depends on need to be registered in the constructor:
```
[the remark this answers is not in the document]
gfontorbe: I think that is a wording problem on my part. But basically in services you initialise only the services you need from a list of all services that is passed as an argument to the constructor.

This is from the DefaultCompletionProvider for example:

```Typescript
protected readonly scopeProvider: ScopeProvider;
    protected readonly ruleInterpreter: RuleInterpreter;
    protected readonly grammar: ast.Grammar;

    constructor(services: LangiumServices) {
        this.scopeProvider = services.references.ScopeProvider;
        this.ruleInterpreter = services.lsp.completion.RuleInterpreter;
        this.grammar = services.Grammar;
    }
```

Seems that way for most (all?) services

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+        this.serviceN = services.Group.SubGroup.ServiceN;
+    }
+    /* service logic */
+}
+```
+
+### Overriding and Extending Services
+Sometimes, you would want to change the behavior of a service or add to its functionality. Thanks to the *inversion of control* via *dependency injection*, this can be achieved in one place without having to modify every piece of code that depends on the service to be overridden or extended.
+
+Let's say that you want to fully replace a service named `MyService` which implements `MyServiceInterface` interface and instantiate an instance of the class `DefaultServiceClass`.
```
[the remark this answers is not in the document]
gfontorbe: It actually *is* a Langium's default service. I will rephrase to go directly to the example.

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+#### LangiumDocumentFactory
+The `LangiumDocumentFactory` service is responsible for creating Langium documents from a `TextDocument`, a `string`, or a `model`.
+
+*Dependencies:* [ServiceRegistry](#serviceregistry)
+#### DocumentBuilder
+The `DocumentBuilder` is responsible for building and updating Langium documents.
+
+*Dependencies:* [Connection](#connection) | [LangiumDocuments](#langiumdocuments) | [IndexManager](#indexmanager) | [ServiceRegistry](#serviceregistry)
+#### TextDocuments
+The `TextDocuments` service manages simple text documents.
```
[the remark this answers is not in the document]
gfontorbe: I went with have an exhaustive list of services. I don't really see the point of removing a couple services from the list.

content/docs/configuration-services.md
gfontorbe: Is that by design? 

I found it quite confusing that the class and interface have the exact same name when I was navigating the code.

content/docs/configuration-services.md
gfontorbe: I will change the sentence

content/docs/configuration-services.md
gfontorbe: I think dealing with the reorganization of the adding and overriding parts, as suggested by Christian will make this look more in place.

content/docs/configuration-services.md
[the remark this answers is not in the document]
sailingKieler: Maybe I was not precise enough, I would just mention the 4-5 most relevant components, since this list will evolve over time.

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+Langium comes pre-packed with a set of services which will handle the most usual features of your language and the LSP. The following gives a brief overview of these default services and their dependencies.
+
+### DefaultSharedModule
+The `DefaultSharedModule` contains services shared between all languages.
+#### ServiceRegistry
+The `ServiceRegistry`service is the core of the service pattern and is responsible for registering and accessing all services.
+#### Connection
+The `Connection` service is responsible for handling the communication between the client and the server.
+#### LangiumDocuments
+The `LangiumDocuments` service manages Langium documents. It is responsible for adding, getting, and removing Langium documents.
```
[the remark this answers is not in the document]
gfontorbe: I think it adds Langium documents to a map of uri and langium documents that are present in the workspace. I will rephrase that.

content/docs/configuration-services.md
msujew: Well they don't have have the exact same name, the interface `MyService` describes the API other services interact with, while the implementation is usually just `DefaultMyService`, since it's a default implementation. I believe that's the common convention in most languages and frameworks, except for .Net, where it would be `IMyService` (interface) and `MyService` (implementation).

content/docs/configuration-services.md
gfontorbe: It is used by more services than the `AstNodeDescriptionProvider`. Basically everytime that a service needs the name of a node it calls the `NameProvider`

content/docs/configuration-services.md
```diff
@@ -0,0 +1,327 @@
...
+```typescript
+export interface LanguageMetaData {
+    languageId: string;
+    fileExtensions: string[];
+    caseInsensitive: boolean;
+}
+```
+
+### LanguageModule
+The `LanguageModule` contains the added and overridden services. See [this section](#customization) for more information.
```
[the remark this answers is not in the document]
gfontorbe: Gone

content/docs/configuration-services.md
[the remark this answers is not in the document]
spoenemann: After rethinking, I would prefer to list three categories of services here:
 * Services that are involved in the document lifecycle (to be documented in a separate chapter)
 * Services for LSP features (list them at the end of this chapter)
 * Utility services (don't need to be mentioned here, comments in the code base should be sufficient)

spoenemann (approved):
Good work, thanks!

PR #23, opened 2022-02-01
Reordered and extended grammar documentation

spoenemann:
* Moved EBNF syntax for terminal rules and "Terminal Fragments" to the end.
 * Moved declaration and entry rule description to the beginning of the "Parser Rules" section.
 * Moved "Cardinalities" from terminal rules to parser rules EBNF syntax.
 * Added "Groups" and "Alternatives" to parser rules EBNF syntax (similar to terminal rules).
 * Moved three larger examples to the end of the "Parser Rules" section ("More Examples").
 * Added more explanation to "Tree-Rewriting Actions".

### DIFF
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -175,7 +175,7 @@ In contrast, an assigned rule call such as `parameter=DeclaredParameter` means t
 #### Unordered Groups
 *Unordered groups are not supported in the current version of Langium. Please refer to [this issue](https://github.com/langium/langium/issues/314) for further information.*
 
-In regular groups, expressions must occur in the exact order as they are declared.
+In regular groups, expressions must occur in the exact order they are declared.
 ```
 Person:
     'person' name=ID age=INT
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -1,7 +1,7 @@
 ---
 title: "The Grammar Language"
 ---
-The grammar language describes the syntax and structure of your language. The [Langium grammar language](https://github.com/langium/langium/blob/main/packages/langium/src/grammar/langium-grammar.langium) is implemented using Langium itself and therefore follows the same syntactic rules as any language created with Langium. The grammar language will define the structure of the *abstract syntax tree* (AST) which in Langium is a collection of *TypeScript types* describing the content of a parsed document and organized hierarchically. The individual nodes of the tree are then represented *at runtime* with JavaScript objects.
+The grammar language describes the syntax and structure of your language. The [Langium grammar language](https://github.com/langium/langium/blob/main/packages/langium/src/grammar/langium-grammar.langium) is implemented using Langium itself and therefore follows the same syntactic rules as any language created with Langium. The grammar language will define the structure of the *abstract syntax tree* (AST) which in Langium is a collection of *TypeScript types* describing the content of a parsed document and organized hierarchically. The individual nodes of the tree are then represented with JavaScript objects at runtime.
 
 In the following, we describe the Langium syntax and document structure.
 ## Language Declaration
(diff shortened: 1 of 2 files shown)


content/docs/grammar-language.md
```diff
@@ -170,82 +76,68 @@ Person returns OtherType:
...
 ```
-In this example, a `Person` can either be a `Student` or a `Teacher`. This grammar is ambiguous because the parser rules `Student` and `Teacher` are identical. The parser will not be able to differentiate between the parser rules for `Student` and `Teacher` when trying to parse a `Person`. 
-Keywords can help removing such ambiguity and guide the parser in defining if a `Student` or `Teacher` needs to be parsed.
-We can add a keyword to the parser rule `Student`, `Teacher`, or to both of them:
+entry Model:
+    (persons+=Person | greetings+=Greeting)*;
 ```
-Student:
-    'student' name=ID;
+In this example, the *entry rule* `Model` defines a group of alternatives. The parser will go through the input document and try to parse `Person` or `Greeting` objects and add them to the `persons` or `greetings` arrays, respectively. The parser reads the token stream until all inputs have been consumed.
```
gfontorbe, line 85: Am I correct that the parser reads all inputs until the end of the document **because** of the cardinality * ?

Wouldn't the parser stop after the first match if we do not have the cardinality * ?

If so then I think it would be good to mention it here.

```suggestion
In this example, the *entry rule* `Model` defines a group of alternatives. The parser will go through the input document and try to parse `Person` or `Greeting` objects and add them to the `persons` or `greetings` arrays, respectively. The parser reads the token stream until all inputs have been consumed thanks to the cardinality `*`.
```

content/docs/grammar-language.md
```diff
@@ -170,82 +76,68 @@ Person returns OtherType:
...
 Person:
-    Student | Teacher;
+    'person' name=ID address=Address;
 ```
-Now the ambiguity is resolved and the parser is able to differentiate between the two parser rules.
+In this example, the rule `Person` must start with the `person` keyword followed by an `ID` token and an instance of the `Address` rule.
 
-Parser rules can have many keywords:
+#### Alternatives
+It is possible to match one of multiple valid options by using the pipe operator `|`. The already mentioned `Model` example specifies to parse either `Person` or `Greeting` inside the parentheses:
```
gfontorbe, line 106: To me it reads like an alternative will always be in parentheses.

```suggestion
It is possible to match one of multiple valid options by using the pipe operator `|`. The already mentioned `Model` example specifies to parse either `Person` or `Greeting`, zero or many times (cardinality *):
```

content/docs/grammar-language.md
```diff
@@ -170,82 +76,68 @@ Person returns OtherType:
...
-    ```
-    paragraph "expression one" /* comment */ "expression two" 3
-    ```
-    The above example will be successfully parsed.
+    Here, the array property `addresses` will accept two expressions matching the terminal rule `STRING`.
 
 3. `?=` is used to assign a **value to a property of type boolean**. The value of the property of type `boolean` is set to `true` if the right part of the assignment is consumed by the parser.
     ```
     Employee:
         'employee' name=ID (remote?='remote')?
```
gfontorbe, line 136: Is there a difference between `(remote?='remote')?` and `remote?='remote'?` ?

They both seem to do what I expect in the LSP, is there some differences to be expected in the AST/CST ?

content/docs/grammar-language.md
```diff
@@ -269,10 +161,21 @@ Hello Sara !
...
+The parser rule `AbstractDefinition` will not create an object of type AbstractDefinition. Instead, it calls either the `Definition` or `DeclaredParameter` parser rule which will be responsible for creating an object of a given type (or call other parser rules if they are unassigned rule calls themselves).
+
+In contrast, an assigned rule call such as `parameter=DeclaredParameter` means that an object is created in the current parser rule and assigns the reult of the `DeclaredParameter` parser rule to the specified property `parameter` of that object.
+
 #### Unordered Groups
-*Unordered groups are not supported in the current version of Langium, please refere to [this issue](https://github.com/langium/langium/issues/314) for further information*
+*Unordered groups are not supported in the current version of Langium. Please refer to [this issue](https://github.com/langium/langium/issues/314) for further information.*
 
-By default, a parser rule has to be implemented in the exact order it is declared.
+In regular groups, expressions must occur in the exact order as they are declared.
```
gfontorbe, line 178:
```suggestion
In regular groups, expressions must occur in the exact order they are declared.
```

gfontorbe (commented):
I think the documentation flows better with the new structure.

Just a handful of picky comments.

content/docs/grammar-language.md
spoenemann: The parser always tries to consume the whole token stream. In case there are syntax errors, it may stop earlier (depending on the error recovery capabilities), but that's independent of the cardinality.

spoenemann: Of course it's a syntax error to write more than one person or greeting if the `* ` is omitted, but the parser will still read the superfluous tokens.

content/docs/grammar-language.md
spoenemann: No it's exactly the same, so the parentheses are not necessary here.

PR #28, opened 2022-02-11
Reworked new configuration chapter

spoenemann:
I did another review round over the new configuration chapter and applied my changes here.

### DIFF
--- a/content/docs/configuration-services.md
+++ b/content/docs/configuration-services.md
@@ -1,34 +1,33 @@
 ---
 title: "Configuration via Services"
 ---
-Langium supports the configuration of most aspects of your language and language server via a set of *Services*. Those services are organized into *Modules*, which are essentially mappings from a service name to its implementation.
 
-We can separate modules into two main categories:
-1. ***Shared modules*** which contain services that are shared across all languages.
-2. ***Language specific modules*** which contain services specific to one language. 
+Langium supports the configuration of most aspects of your language and language server via a set of *services*. Those services are configured by *modules*, which are essentially mappings from a service name to its implementation.
 
-## Shared Modules
-The *shared modules* contain services shared across all languages:
-* The `ServiceRegistry` is responsible for registering and accessing the different services.
-* The `Connection` service sends messages to the client and register message handlers for incoming messages.
-* Services to handle *Langium Documents* and *Text Documents*.
-* Implementations of the `AstReflection` service (one implementation per language) to access the structure of the AST.
+We can separate services and modules into two main categories:
 
-## Language Specific Modules
-The *language specific modules* contain all services which are specific to one language:
+#### Shared Services
+The *shared services* are services that are shared across all Langium languages. In many applications there is only one Langium language, but the overall structure of the services is the same.
+* The `ServiceRegistry` is responsible for registering and accessing the different languages and their services.
+* The `Connection` service is used in a language server context; it sends messages to the client and registers message handlers for incoming messages.
+* The `AstReflection` service provides access the structure of the AST types.
+* Shared services involved in the document lifecycle (future documentation)
+
+#### Language Specific Services
+The *language specific services* are services specific to one Langium language and isolated from other languages.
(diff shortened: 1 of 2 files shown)


content/docs/configuration-services.md
```diff
@@ -1,34 +1,33 @@
...
+The *shared services* are services that are shared across all Langium languages. In many applications there is only one Langium language, but the overall structure of the services is the same.
+* The `ServiceRegistry` is responsible for registering and accessing the different languages and their services.
+* The `Connection` service is used in a language server context; it sends messages to the client and registers message handlers for incoming messages.
+* The `AstReflection` service provides access the structure of the AST types.
+* Shared services involved in the document lifecycle (future documentation)
+
+#### Language Specific Services
+The *language specific services* are services specific to one Langium language and isolated from other languages.
 * Services for [LSP features](#language-server-protocol)
 * Services involved in the document lifecycle (future documentation)
```
gfontorbe, line 19: Just after a quick look at the 'createModule functions', aren't all document services shared?

spoenemann: Services such as `Linker` and `DocumentValidator` are language-specific and also involved in the document lifecycle.

content/docs/configuration-services.md
```diff
@@ -119,12 +120,12 @@ export const ArithmeticsModule: Module<ArithmeticsServices, PartialLangiumServic
...
 The services which `ServiceClass` depends on need to be registered in the constructor:
 ```Typescript
-export class ServiceClass{
-    readonly serviceOne: ServiceOne;
-    readonly serviceTwo: ServiceTwo;
-    readonly serviceN: ServiceN;
+export class ServiceClass {
+    private readonly serviceOne: ServiceOne;
+    private readonly serviceTwo: ServiceTwo;
+    private readonly serviceN: ServiceN;
```
msujew, line 126: Having `private` on service fields makes it difficult to extend the service class. Even we don't expect end users to create a service class inheritance structure, I believe it to be good practice to have them in the doc as `protected`.

content/docs/configuration-services.md
```diff
@@ -181,13 +182,13 @@ export const ArithmeticsModule: Module<ArithmeticsServices, PartialLangiumServic
 If you want to modify aspects of the Language Server, this section will help you find the relevant service for handling a given LSP request.
 
 #### CompletionProvider
-The `CompletionProvider` service is responsible for handling a [LSP Completion Request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion) at a given cursor position. When a *LSP Completion Request* is submitted by the client to the server, the `CompletionProvider` will create a `CompletionList` of all possible `CompletionItem` to be presented in the editor. The `CompletionProvider` service computes a new `CompletionList` after every typing.
+The `CompletionProvider` service is responsible for handling a [Completion Request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion) at a given cursor position. When a *Completion Request* is submitted by the client to the server, the `CompletionProvider` will create a `CompletionList` of all possible `CompletionItem` to be presented in the editor. The `CompletionProvider` service computes a new `CompletionList` after every typing.
```
msujew, line 185:
```suggestion
The `CompletionProvider` service is responsible for handling a [Completion Request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion) at a given cursor position. When a *Completion Request* is submitted by the client to the server, the `CompletionProvider` will create a `CompletionList` of all possible `CompletionItem` to be presented in the editor. The `CompletionProvider` service computes a new `CompletionList` after each keystroke.
```

content/docs/configuration-services.md
spoenemann: I agree from the perspective of the library / framework code. In user code, I tend to make things private by default because I can simply change that in case I need another visibility. And I would like to not encourage using inheritance in user code. The only reason we use it in Langium is to override default functionality.

PR #31, opened 2022-02-25
Update documentation on guard conditions

msujew:
Given that https://github.com/langium/langium/pull/422 has been merged, we should update the documentation on guard conditions.

### DIFF
--- a/content/docs/grammar-language.md
+++ b/content/docs/grammar-language.md
@@ -294,7 +294,7 @@ TechnicalStaff:
 
 Fragment rules are not part of the AST and will therefore never create an object, instead they can be understood as being textually inserted where they are referenced.
 
-### Guarded Rules
+### Guard Conditions
 It may be useful to group parser rules with small variations inside of a single parser rule. Given the following example:
 ```
 entry Model:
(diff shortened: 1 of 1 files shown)


content/docs/grammar-language.md
```diff
@@ -312,25 +312,37 @@ Element:
...
 ```
-`Element` has the guard `isRoot`, which will determine which of the two alternatives should be matched by the parser. 
-If `isRoot` is set to `true`, then the first alternative is matched. Otherwise, the second alternative will apply. 
+`Element` has the guard `isRoot`, which will determine whether the optional group containing the `isPublic` property is allowed to be parsed.
+
 The *entry rule* `Model` sets the value of `isRoot` to `true` with `element+=Element<true>`, while `isRoot` is set to `false` inside of the `Element<isRoot>` parser rule with `elements+=Element<false>`.
 
+In general, a guard condition on a group decides whether the parser is allowed to parse the group or not depending on the result of the evaluated condition. Logical operations can be applied, such as `&` (and), `|` (or) and `!` (not) to fine-tune the exact conditions in which the group is supposed to be parsed.
+
+Additionally, guard conditions can also be used inside of alternatives. See the following example:
```
gfontorbe, line 332: Just to be sure I am clear about the syntax, let's consider:
```
Element<isRoot>:
	(<isRoot> 'root' | 'element') name=ID '{'
		elements+=Element<false>*
	'}';
```

Would this be equivalent to the example you give in the documentation?

gfontorbe (commented):
Thanks @msujew , just a quick question about the syntax below

content/docs/grammar-language.md
msujew: It wouldn't there's a slight difference:

`(<isRoot> 'root' | 'element')` __always__ allows to parse `element`, while it allows `root` if `isRoot=true`, while `(<isRoot> 'root' | <!isRoot> 'element')` makes it __mutually exclusive__. Only one is possible at any given moment.

PR #33, opened 2022-03-04
Update getting started guide to 0.3

msujew:
Includes minor changes from 0.3 which have to be reflected in the documentation.

### DIFF
--- a/content/docs/getting-started.md
+++ b/content/docs/getting-started.md
@@ -72,12 +72,12 @@ For a full explanation of all terms we use throughout Langium, please refer to o
 Here's the grammar that parses the previous text snippet:
 
 ```
-grammar HelloWorld hidden(WS)
+grammar HelloWorld
 
-terminal WS: /\s+/;
+hidden terminal WS: /\s+/;
 terminal ID: /[_a-zA-Z][\w_]*/;
 
-Model: (persons+=Person | greetings+=Greeting)*;
+entry Model: (persons+=Person | greetings+=Greeting)*;
 
 Person:
     'person' name=ID;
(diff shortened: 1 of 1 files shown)


content/docs/getting-started.md
```diff
@@ -72,12 +72,12 @@ For a full explanation of all terms we use throughout Langium, please refer to o
 Here's the grammar that parses the previous text snippet:
 
 ```
```
gfontorbe, line 74: The grammar that is generated by `yo langium` is actually a bit different, terminals are at the end, there are terminal rules for INT STRING and comments, and type for the reference to Person:
```
grammar HelloWorld

entry Model:
    (persons+=Person | greetings+=Greeting)*;

Person:
    'person' name=ID;

Greeting:
    'Hello' person=[Person:ID] '!';

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;
terminal INT returns number: /[0-9]+/;
terminal STRING: /"[^"]*"|'[^']*'/;

hidden terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
hidden terminal SL_COMMENT: /\/\/[^\n\r]*/;
```

gfontorbe (commented):
The grammar file is a bit different on a newly generated project. Apart from that, LGTM.

content/docs/getting-started.md
msujew: Given that these rules are basically only included to have reference points how these terminals would usually look like, I would not try to include them in the getting started page. The point of this page is to explain how one would get started with a Langium project. Throwing unused terminals in there likely just leads to confusion. @spoenemann WDYT?

spoenemann: Yes, let's omit unused terminals to avoid confusion.

spoenemann (approved):

PR #38, opened 2022-03-17
Added more details to document lifecycle

spoenemann:
Are the new explanations understandable?

### DIFF
--- a/content/docs/document-lifecycle.md
+++ b/content/docs/document-lifecycle.md
@@ -1,7 +1,9 @@
 ---
 title: 'Document Lifecycle'
-weight: 300
+weight: 400
 ---
+
+`LangiumDocument` is the central data structure in Langium that represents a text file of your DSL. Its main purpose is to hold the parsed Abstract Syntax Tree (AST) plus additional information derived from it. After its creation, a `LangiumDocument` must be "built" before it can be used in any way. The service responsible for building documents is called `DocumentBuilder`.
 
 A `LangiumDocument` goes through seven different states during its lifecycle:
 1. `Parsed` when an AST has been generated from the content of the document.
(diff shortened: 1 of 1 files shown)


content/docs/document-lifecycle.md
```diff
@@ -12,40 +14,44 @@ A `LangiumDocument` goes through seven different states during its lifecycle:
 6. `Validated` when the document has been validated by the `DocumentValidator`.
 7. `Changed` when the document has been modified.
 
-The following diagram depicts how the `DocumentBuilder` processes `LangiumDocument`s depending on their state. More details about each step of the lifecycle can be found below.
+State 1 is the initial state after creation of a document, and states 2 to 6 are part of its build process. The following diagram depicts how the `DocumentBuilder` processes `LangiumDocument`s depending on their state. More details about each step of the lifecycle can be found below.
```
gfontorbe, line 17: Maybe add that state 7 is outside of the build process?

content/docs/document-lifecycle.md
```diff
@@ -1,8 +1,10 @@
 ---
 title: 'Document Lifecycle'
-weight: 300
+weight: 400
 ---
 
+`LangiumDocument` is the central data structure in Langium that represents a text file of your DSL. Its main purpose is to hold the parsed Abstract Syntax Tree (AST) plus additional information derived from it. After its creation, a `LangiumDocument` must be "built" before it can be used in any way. The service responsible for building documents is called `DocumentBuilder`.
```
gfontorbe, line 6: Might be a bit too much of a niche for this part of the documentation but should we mention that `LangiumDocument` can also be generated programatically and don't have to necessarily come from a text file?

gfontorbe, line 6: Might be a bit too much of a niche for this part of the documentation but should we mention that `LangiumDocument` can also be generated programatically and don't have to necessarily come from a text file?

content/docs/document-lifecycle.md
```diff
@@ -57,13 +63,15 @@ B --> D(<b>AstNodeLocator</b><br><br><i>gets the path of an AstNode</i>)
...
 
-After all pre-processing steps have been completed, the document's state is set to `Processed`.
+The default implementation of the `ScopeComputation` service attaches the description of every symbol to its direct container. This means that the container holds information about which named nodes are nested inside of it. You can override this default behavior to change the position where a symbol is reachable, or to change the name by which it can be referenced. It is even possible to associate the same symbol to multiple container nodes, possibly with different names, to control precisely where and how references to it can be resolved. However, keep in mind that you cannot access any cross-references in this phase. More complex, context-dependent scope mechanisms can be implemented in the `ScopeProvider` (see below).
```
gfontorbe, line 70: This might be difficult to follow without a concrete example (maybe we could link to the domainmodel example which implements a custom ScopeComputation?)

gfontorbe, line 70: This might be difficult to follow without a concrete example (maybe we could link to the domainmodel example which implements a custom ScopeComputation?)

gfontorbe (commented):
This is much more detailed but still written in a way that is easy enough to follow and understand imo. Just a few comments here and there but nothing major.

content/docs/document-lifecycle.md
spoenemann: I would cover that in a separate guide about built-in libraries.

PR #45, opened 2022-05-20
Ast types doc

gfontorbe:
Draft for the documentation about AST Types.

Every comment/suggestion is welcome!


hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
 weight: 200
 ---
 
-Coming soon...
+When AST nodes are created during the parsing of a document, they are given a type. The language grammar dictates the shape of those types and how they might be related to each other. There are two ways by which Langium derives AST types from the grammar, by **[inference](#inferred-types)** and by **[declaration](#declared-types)**.
+
+*Inference* is the default behavior in Langium. During the generation of the AST types, Langium infers the possible types directly from the grammar rules. While this a powerful approach for simple languages and prototypes, it is not recommended for more mature languages since minimal changes in the grammar can easily lead to breaking changes.
```
dhuebner, line 8: > While this a powerful approach

Feels like something is missing here

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+
+Explicitly declaring type aliases in the grammar is achieved with the keyword `type`:
+```
+type X = A | B;
+
+// generates:
+type X = A | B;
+```
+
+Please note that it is not allowed to use an alias type as a return type in a parser rule. The following syntax is invalid:
```
dhuebner, line 271: Would be helpful to see a valid example.

hugo/content/docs/ast-types.md
montymxb: While this **is** a powerful approach...

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+
+## Inferred Types
+*Inferred types* result from letting Langium infer the types of AST nodes from the grammar rules. Let's have a look at how various rules shape the AST:
+
+### Parser Rules
+The simplest way to write a parser rule is as follows:
+```
+X: name=ID;
+```
+With this syntax, Langium will **infer** the type of the AST node to be generated when parsing the rule. By convention, the type of the AST node will be named after the name of the rule, resulting in this **TypeScript's interface** in the AST:
```
montymxb, line 22: resulting in this **Typescript interface** in the AST:

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+```
+interface X extends AstNode {
+    name: string
+}
+```
+It is however possible to control the naming of the interface by using the following syntax:
```
montymxb, line 28: It is **also** possible...

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+This result in the creation of a single interface in the AST 'merging' the two parser rules with non-common properties made optional:
+```
+interface MyType extends AstNode {
+    count?: number
+    name: string
+}
+```
+
+### Terminal Rules
+Terminal rules are linked to built-in types in the AST. They do not result in AST types on their own but determine the type of properties in AST types inferred from parser rule:
```
montymxb, line 56: ...inferred from **the** parser rule:

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+// generates:
+interface X extends AstNode {
+    name: string
+    count: number
+}
+```
+The property `name` is of type `string` because the terminal rule `ID` is linked to the built-in type `string`, and the property `count` is of type `number` because the terminal rule `INT` is linked to the built-in type `number`.
+
+### Data type rules
+Data type rules are similar to terminal rules in the sense that they determine the type of properties in AST types inferred from a parser rules. However, they lead to the creation of type aliases for built-in types in the AST:
```
montymxb, line 72: ... in AST types inferred from **~a~** parser rules.

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+    numbers: Array<number>
+    isValid: boolean
+}
+```
+
+The right side of an assignment can be any of the following:
+* A terminal rule or a data type rule, which results in the type of the property to be a built-in type.
+* A parser rule, which results is the type of the property to be the type of the parser rule.
+* A cross-reference, which results in the type of the property to be a *Reference* to the type of the cross-reference.
+* An alternative of all the above, which results in the type of the property to be a type union of all the types in the alternative.
```
montymxb, line 108: Could just say `An alternative, which results...`, and drop the `of all the above` part. As an alternative could contain an alternative itself, which is not listed above.

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+}
+
+interface Y extends AstNode {
+    crossValue: Reference<X>
+    alt: number | X | Reference<X>
+}
+```
+
+### Unassigned Rule Calls
+A parser rule does not necessarily need to have assignment, but can also only contain *unassigned rule calls*. These kind of rules can be used to change the types' hierarchy.
```
montymxb, line 127: A parser rule does not necessarily need to have **assignments**. **It may** also **contain only** *unassigned rule calls*.

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+}
+
+interface B extends AstNode {
+    name: string
+    count: number
+}
+```
+
+### Simple Actions
+Actions can be used to infer the type of the AST node **inside** of a parser rule. This can be viewed as *syntactic sugar* and can increase readability of the grammar.
```
montymxb, line 148: **These** can be viewed as...

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,338 @@ title: "AST Type Inference"
...
+}
+
+interface B extends AstNode {
+    name: string
+    count: number
+}
+```
+
+### Assigned actions
+Actions can also be used to control the structure of AST node types. This is a more advanced topic, and we recommend getting familiar with the rest of the documentation before diving into this section.
```
montymxb, line 173: This is a more advanced topic, **so** we recommend...

montymxb (commented):
Was reading through the document again for the purpose of learning more about how AST types are inferred/declared, quite helpful there 🙂 . I also made some more suggestions for small edits, but all around looks good.

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+interface X extends AstNode {
+    name: string
+}
+```
+
+### Assignments
+There are three available kinds of [assignments](../grammar-language/#assignments) in a parser rule:
+
+1. `=` for assigning a **single value** to a property, resulting in the property's type to be derived from the right side of the assignment.
+2. `+=` for assigning **multiple values** to a property, resulting in the property's type to be an array of the right side of the assignment.
```
msujew, line 90:
```suggestion
1. `=` for assigning a **single value** to a property, resulting in the property's type to be derived from the right hand side of the assignment.
2. `+=` for assigning **multiple values** to a property, resulting in the property's type to be an array of the right hand side of the assignment.
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+interface B extends AstNode {
+    name: string
+    count: number
+}
+```
+
+### Assigned actions
+Actions can also be used to control the structure of AST node types. This is a more advanced topic, so we recommend getting familiar with the rest of the documentation before diving into this section.
+
+Let's consider two different grammars derived from the [Arithmetics example](https://github.com/langium/langium/blob/main/examples/arithmetics/src/language-server/arithmetics.langium). These grammars are designed to parse a document containing a single definition comprised of a name and an expression assignment, with an expression being an indefinite length of additions or a numerical value.
```
msujew, line 175:
```suggestion
Let's consider two different grammars derived from the [Arithmetics example](https://github.com/langium/langium/blob/main/examples/arithmetics/src/language-server/arithmetics.langium). These grammars are designed to parse a document containing a single definition comprised of a name and an expression assignment, with an expression being an any amount of additions or a single numerical value.
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+The first one does not use assigned actions:
+```
+Definition: 
+    'def' name=ID ':' expr=Expression;
+
+Expression:
+    '(' Addition ')' | value=NUMBER;
+
+Addition infers Expression:
+    left=Expression ('+' right=Expression)*;
```
msujew, line 186: I believe this grammar doesn't explain the issue well enough, since it's quite restrictive (it needs parenthesis to parse) and does some weird stuff with it's addition rule: When repeatedly parsing `right=Expression`, we override the previous `right` assignment, effectively "deleting" the previous value. I would recommend this grammar instead:

```suggestion
Definition: 
    'def' name=ID ':' expr=Expression;

Expression:
    Addition;

Addition infers Expression:
    left=Value ('+' right=Expression)?;
    
Primary infers Expression:
    '(' Expression ')' | {Literal} value=NUMBER;
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+expr((expr)) --> left((left))
+expr --> right((right))
+left --> left_left((left))
+left --> left_right((right))
+right --> right_right((right))
+left_left --> left_left_v{1}
+left_right --> left_right_{2}
+right_right --> right_right_v{3}
+{{</mermaid>}}
+We can see that the nested `right -> right` property in the tree is counter-intuitive and we would like to remove one level of nesting from the tree.
```
msujew, line 200:
```suggestion
When parsing a document containing `def x: (1 + 2) + 3`, this is the shape of the AST node:
{{<mermaid>}}
graph TD;
expr((expr)) --> left((left))
expr --> right((right))
left --> left_left((left))
left --> left_right((right))
right --> right_left((left))
left_left --> left_left_v{1}
left_right --> left_right_{2}
right_left --> right_left_v{3}
{{</mermaid>}}
We can see that the nested `right -> left` property in the tree is counter-intuitive and we would like to remove one level of nesting from the tree. 
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+left --> left_left((left))
+left --> left_right((right))
+right --> right_right((right))
+left_left --> left_left_v{1}
+left_right --> left_right_{2}
+right_right --> right_right_v{3}
+{{</mermaid>}}
+We can see that the nested `right -> right` property in the tree is counter-intuitive and we would like to remove one level of nesting from the tree. 
+
+This can be done by reworking the grammar and adding an assigned action:
```
msujew, line 202:
```suggestion
This can be done by refactoring the grammar and adding an assigned action:
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+```
+Definition: 
+    'def' name=ID ':' expr=Addition ';';
+
+Expression:
+    '(' Addition ')' | value=NUMBER;
+
+Addition infers Expression:
+    Expression ({infer Addition.left=current} '+' right=Expression)*;
```
msujew, line 211: To continue with the improvement from above:

```suggestion
Definition: 
    'def' name=ID ':' expr=Addition ';';

Expression:
    Addition;

Addition infers Expression:
    Primary ({infer Addition.left=current} '+' right=Primary)*;
    
Primary infers Expression:
    '(' Expression ')' | {Literal} value=NUMBER;
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+expr --> right((right))
+left --> left_left((left))
+left --> left_right((right))
+right --> right_v{3}
+left_left --> left_left_v{1}
+left_right --> left_right_{2}
+{{</mermaid>}}
+
+## Declared Types
+Because type inference takes into account every entity of a parser rule, even the smallest changes can update the inferred types. This can lead to an unexpected type system and incorrect behavior of services that depend on it. To minimize the risk of introducing breaking changes when modifying the grammar, we recommend to use *declared types*. This is especially true for more mature and complex languages, where a stable type system is key and breaking changes introduced by inferred types can be hard to detect. Declared types allow the user to **fix** the type of the parser rules and rely on the power of validation errors to detect breaking changes.
```
msujew, line 226:
```suggestion
Because type inference takes into account every entity of a parser rule, even the smallest changes can update the inferred types. This can lead to an unexpected meta model and incorrect behavior of services that depend on it. To minimize the risk of introducing breaking changes when modifying the grammar, we recommend to use *declared types*. This is especially true for more mature and complex languages, where a stable type system is key and breaking changes introduced by inferred types can be hard to detect. Declared types allow the user to **fix** the type of the parser rules and rely on validation errors to detect breaking changes.
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+X returns MyType: name=ID;
+Y returns MyType: name=ID count=INT;
+```
+We now explicitly declare `MyType` directly in the grammar with the keyword `interface`. The parser rules `X` and `Y` creating AST nodes of type `MyType` need to explicitly declare the type of the AST node they create with the keyword `returns`.
+
+Contrary to [inferred types](#inferred-types), all properties must be explicitly declared in order to be valid inside of a parser rule. The following syntax:
+```
+Z returns MyType: name=ID age=INT;
+```
+will throw the following validation error `A property 'age' is not expected` because the declaration of `MyType` does not include the property `age`. In short, *declared types* add a layer of safety via validation to the grammar that prevents mismatches between the generated AST types and what the user expects.
```
msujew, line 248:
```suggestion
will show the following validation error `A property 'age' is not expected` because the declaration of `MyType` does not include the property `age`. In short, *declared types* add a layer of safety via validation to the grammar that prevents mismatches between the expected meta model types and the shape of the parsed AST nodes.
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+
+A declared type can also extend another declared type:
```
msujew, line 250:
```suggestion
A declared type can also extend types, such as other declared types or types inferred from parser rules:
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+}
+
+interface MyOtherType extends MyType {
+    count: number
+}
+
+Y returns MyOtherType: name=ID count=INT;
+```
+
+Explicitly declaring type aliases in the grammar is achieved with the keyword `type`:
```
msujew, line 263:
```suggestion
Explicitly declaring union types in the grammar is achieved with the keyword `type`:
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+```
+type X = A | B;
+
+// generates:
+type X = A | B;
+```
+
+Please note that it is not allowed to use an alias type as a return type in a parser rule. The following syntax is invalid:
```
msujew, line 271:
```suggestion
Please note that you cannot use union types as a return type of a parser rule. The following is invalid:
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+```
+type X = A | B;
+
+Y returns X: name=ID;
+```
+
+Using `returns` without explicitly declaring the type of the AST node is not allowed and results in a validation error.
```
msujew, line 278:
```suggestion
Using `returns` always expects a reference to an already existing type. To create a new type for your rule, use the `infers` keyword.
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+
+X returns A: reference=[B:ID] array+=Y (array+=Y)* alternative=(Y | Z);
+
+Y returns B: 'Y' name=ID;
+
+Z returns C: 'Z' name=ID count=INT;
+```
+
+### Actions
+Actions referring to a declared type have the following syntax:
```
msujew, line 306:
```suggestion
Actions referring to a declared type can be seen in the following:
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+interface B {
+    name: string
+    count: number
+}
+
+X: 
+    {A} 'A' name=ID 
+  | {B} 'B' name=ID count=INT;
+```
+Note the absence of the keyword `infer`, contrary to [actions inferring types](#simple-actions).
```
msujew, line 321:
```suggestion
Note the absence of the keyword `infer` compared to [actions which infer a type](#simple-actions).
```

hugo/content/docs/ast-types.md
```diff
@@ -3,4 +3,340 @@ title: "AST Type Inference"
...
+
+## Refactoring Dummy Rules
```
msujew, line 323: I'm not too sure about this section in general. It sounds really forced in a way and is probably only for people coming from Xtext? If that's the case, we should make the reference here explicit, and if it's not, then we should probably remove this section.

msujew (commented):
In general, I believe we should call this page `Meta Model Inference`, since using AST in this context is highly ambiguous. @sailingKieler @spoenemann Do you have anything against that?

hugo/content/docs/ast-types.md
gfontorbe: Seems like this is actually not enforced anymore. I had a check this morning and it was not returning a validation error.

msujew: Well, yes and no. Ideally, it should work like this:

```
type X = A | B;

ARule returns A: ...;
BRule returns B: ...; 
XRule returns X: ARule | BRule; // Valid

YRule returns X: name=ID; // Invalid
```

But you're right, it's anyway incorrect in my suggestion. We'll have to think about this a bit. Do you mind putting this part in a comment as long as we don't have a decision on it?

hugo/content/docs/ast-types.md
gfontorbe: It is actually now possible to return an alias type

hugo/content/docs/ast-types.md
gfontorbe: I'll leave it there for now and maybe we should think about having a documentation page about migrating from XText.

msujew (approved):
Thanks a lot for the work on this, LGTM!

PR #46, opened 2022-05-24
Restructure website to use npm workspaces

msujew:
Creates a new workspace `website` which contains the documentation and landing page. Enables other workspaces containing other content.


package.json
```diff
@@ -1,26 +1,21 @@
...
-    "build:tailwind":     "NODE_ENV=production  tailwindcss -i ./tailwind/style.css -o ./hugo/static/css/style.css -c ./tailwind/tailwind.config.js --minify",
-    "build:tailwind:dev": "NODE_ENV=development tailwindcss -i ./tailwind/style.css -o ./hugo/static/css/style.css -c ./tailwind/tailwind.config.js --watch "
-  },
-  "devDependencies": {
-    "concurrently": "^6.2.2",
-    "hugo-extended": "^0.89.4",
-    "live-server": "^1.1.0",
-    "tailwindcss": "^2.2.16"
+    "name": "langium-website-project",
+    "version": "0.0.0",
```
kaisalmen, line 3: FYI, you can remove the version if it is anyhow a private project. It is not mandatory to keep I found out last week.

package.json
```diff
@@ -1,26 +1,21 @@
...
+    "description": "Source code for langium.org",
+    "author": "TypeFox",
+    "license": "MIT",
+    "private": true,
+    "scripts": {
+      "dev":        "concurrently \"npm run dev --workspace website\"",
```
kaisalmen, line 9: Why do you need concurrently here? If not needed the `devDependencies` could be removed in parent.

kaisalmen (approved):
Comments are non-blocking.

package.json
msujew: We will need to run `npm run dev --workspace <example-language>` concurrently with the already existing build scripts, since they are blocking. So we can't use `&&` in there.

msujew:
@kaisalmen Care to take another look? I restructured it again, since we didn't actually need to package hugo and tailwind into one single package. Instead they are now separate packages.

package.json
```diff
@@ -1,26 +1,25 @@
...
+      "watch":         "concurrently -c gray,blue -k -n hugo,tailwind \"npm run watch        --workspace hugo\" \"npm run watch --workspace tailwind\"",
+      "watch:gitpod":  "concurrently -c gray,blue -k -n hugo,tailwind \"npm run watch:gitpod --workspace hugo\" \"npm run watch --workspace tailwind\"",
+      "build": "npm run clean && npm run build --workspaces",
+      "clean": "rimraf public"
+    },
+    "devDependencies": {
+      "concurrently": "^6.2.2",
+      "cross-env": "^7.0.3",
+      "hugo-extended": "^0.89.4",
+      "rimraf": "^3.0.2",
```
kaisalmen, line 17: FYI, `rimraf` brings in tons of old dependecies. I switched to `shx` for cross-platform support. You could do the same with it like this: `npx shx rm -rf public`

kaisalmen (approved):
Yeah, seems like the better approach. Comment does not block, but may be useful info.

Issue #59, opened 2022-07-21
Insert the right logo svg to docu header


KatharinaBrinkhaus:
Currently on the documentation page there is a simple headline next to the icon not the correct logo.

PR #61, opened 2022-07-22
Image inserted, replaced with old one.

sakibian:
As mentioned in the header file, assets image file(langium_logo_w_nib.svg) is replaced with a new one(langium_logo.svg) which provided by @[KatharinaBrinkhaus](https://github.com/KatharinaBrinkhaus) on issue. #59


msujew (commented):
Hey @sakibian, thanks for contributing!

However, I think you picked the wrong entry point to address the issue. The `langium-header.html` partial isn't responsible for the documentation, but for the landing page. The file you're looking for is likely [this one](https://github.com/langium/langium-website/blob/207c1bd238200af7faaf2218f0c8ded1f778486e/hugo/themes/hugo-geekdoc/layouts/partials/site-header.html#L15-L20).

Also, we already have the svg logo available in [here](https://github.com/langium/langium-website/blob/main/hugo/static/assets/langium_logo_w_nib.svg), please don't duplicate it.

sakibian:
@msujew my bad I thought I understood the layout but I was wrong, thanks for pointing it out, cheers.

PR #62, opened 2022-07-23
extra space added to svg heart icon

KaptainCS3:
target the svg element using class name and added margin left space


montymxb (approved):
Cool! This will work, thanks for contribution!

KaptainCS3:
it's my pleasure

PR #63, opened 2022-08-08
Guide on code bundling

msujew:
Explains why it's important to bundle your code even in a node.js context and shows a few examples, how one can easily bundle:

1. Language server and vscode extension using esbuild and webpack
2. Language server in a web worker using esbuild


spoenemann (approved):
Great! Two thoughts on the docs structure:
 - With the new "Guides" section, we can remove the empty "Tutorials" chapter
 - I think we can now remove the _WORK IN PROGRESS_ message on the documentation entry page as the docs are becoming more mature.

PR #64, opened 2022-08-10
Switch to monaco-editor-wrapper and clean-up

kaisalmen:
Typescript is no longer needed in the showcase, It is basically the HMTL page and langium-statemachine-dsl that is bundled into the worker


kaisalmen:
Hey @montymxb I now put this branch on top of #69 

It removes the need for a separate workspace packages, but brings the devDependencies to the hugo package. Let me know what you think?

kaisalmen:
If you merge the other branch, I will quickly rebase this one on main and then the history (above) is only focused on these changes.

kaisalmen:
Rebase is done

kaisalmen:
@montymxb I have updated this branch to the final version `1.0.0` of `monaco-editor-wrapper`. This can be merged in if you like. The supplied bundle is still used, because we statically link the JavaScript file(s).

Is it possible to used es modules and have some build process in Hugo (I am not familiar enough with Hugo)? But nevertheless, we can just use this approach.

montymxb (approved):
The netlify preview finally kicked in, and it looks good! This simplifies the build process without the need for an extra copy from the showcase directory.

montymxb:
@kaisalmen Hugo has ESBuild support built in, if that works for us. https://gohugo.io/hugo-pipes/js/

PR #66, opened 2022-08-11
Updated documentation about scope computation

spoenemann:
See langium/langium#622 - this has to be reflected in the documentation.


msujew (approved):
LGTM 👍

PR #69, opened 2022-09-01
Build Showcase Page with Hugo Layouts

### LINKED ISSUE #55: Mobile menu does not work on statemachine editor page
The [statemachine editor page](https://langium.org/showcase/statemachine/) is missing the code for the mobile menu.

The html source of that page is currently hard-coded and does not use the predefined templates. The missing mobile menu could be fixed by using a template that includes it, but feel free to fix it directly and create a new issue for using the templates.

### LINKED ISSUE #59: Insert the right logo svg to docu header

montymxb:
This closes #55 & closes #59, and adds several QOL fixes to showcase generation:
- changes build to copy tailwind & showcase assets into **hugo/static** instead of **public**. This is more appropriate than copying directly into public, since things can be overridden there by hugo (this was previously happening with the showcase when using 'watch'). Anything added into static will be served as-is, and will correctly trigger hugo to rebuild
- fixes issues with mobile menu not working for showcase
- fixes issues with mobile footer behaving strangely for showcase
- generalizes the implementation of a showcase page. Leaving the script implementation up to the dev. This can be further abstracted later.
- adds our full logo to the geekDoc config, and modifies the geekDoc templates slightly to display it as intended in our docs


montymxb:
Additional note, there has been a repeat bug with strange build results showing up in the Netlify preview; which also happened here as well. Turns out it was related to our tailwind config, which was purging css based on the contents of **public** and a couple **hugo** folders. Since most of us have already built the site locally at least once, tailwind is able to put together the correct css using **public**. On netlify, for a fresh build without a previous public folder, tailwind just strips out a whole bunch of css that wasn't found.

The fix is included here. Instead of depending on the contents of public, and only partially on hugo, we fully check hugo's layouts, contents, and static files to determine what css to include. Since everything is already there, even in template form, it works out quite nicely and we get what we expect. There may be some edge cases due the way hugo can process and modify the templates, like adding in extra classes to elements. But so far I haven't observed any of these issues in our deployment.

spoenemann (approved):
Thanks!

hugo/content/showcase/domainmodel.html
```diff
@@ -1,11 +1,14 @@
 ---
```
Lotes, line 1: Why the rename? Any advantage or consequence?

montymxb: Changed from markdown to html so allow literal html content in this file. The old markdown file wasn't actually being used as well, as it was overridden by the prebuilt showcase contents on build.

This lets us add in the direct script tag that let us choose the language, syntax highlighting, example file contents, and anything else we need to demonstrate it. More granular control over the individual showcase, without having to define all the surrounding boilerplate and page structure.

Lotes: nice

Lotes (approved):
Very nice solution of splitting showcase-specific parts to their own file.

PR #71, opened 2022-09-16
Fix Footer Issues & Missing Font for Showcase

montymxb:
This addresses some issues that are visible on the current site regarding the footer and showcase page, these are:
- footer elements are not centered
- footer does not compress well for mobile view
- header & footer font is overriden by monaco for the showcase page
The following gif shows these issues:
a2

And this one demonstrates these issues fixed in this PR.
langiumwebsite


spoenemann (approved):
Thanks!

PR #72, opened 2022-09-20
Playground

Lotes:
Adds two editors (left: Langium grammar, right: the editor based on the left Grammar)
+ language server for both sides
+ syntax highlighting for both sides


spoenemann:
When I open this in the deploy preview, the right editor does not work and the browser console prints a few errors.

Could you add a "Playground" entry in the main menu / title bar (after "Showcase")?

Lotes:
An error view and header title is also missing. Menu entry I will add...

The preview is not working. I hoped it is a problem with the provider. I think @montymxb mentioned some problems with the cache. But I am yet not sure. Would be great to find some certainity here...

spoenemann:
Thanks for the fix @msujew!

This looks great @Lotes!

I'd propose to separate this from the Showcase:
 - Change path to `/playground/` (instead of `/showcase/playground/`)
 - Remove link from Showcase page
 - Add Playground link to menu bar as mentioned above
Rationale: Many will find it hard to understand the different abstraction levels of the grammar language vs. applications of it. We should highlight the Playground as a separate feature that shows how you can use Langium, while the Showcase demonstrates specific applications built with Langium.

How do you envision the "error view"? Do you have a sketch?

Lotes:
Added a shadow for the right editor when the left grammar is not correct

spoenemann:
Thanks! I noticed that the overlay is not always removed when the editor turns back into a valid state. Try adding a single character `a` at the end of the grammar, then remove it again.

It would be great to have a loading indicator, for example at the top of the right editor, that is made visible after every keystroke in the left editor and removed when the update of the right editor is finished.

spoenemann (approved):
Thank you, I think this is great for a start. Could you change the initial example to a smaller one like hello-world, and then merge this? We can add further improvements in follow-up PRs.

PR #73, opened 2022-09-23
Adds Details about Negated Tokens in EBNF

montymxb:
Related to an issue opened on Langium, https://github.com/langium/langium/issues/681. Negated tokens don't work in Langium the same way that they do in Xtext. We may address this, but for the sake of transparency, I've updated the docs to make this incompatibility clear.

If we rectify this at a later time, we can update the docs to reflect this.

### DIFF
--- a/hugo/content/docs/grammar-language.md
+++ b/hugo/content/docs/grammar-language.md
@@ -487,24 +487,19 @@ terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
 ``` 
 
 #### Negated Token
-It is possible to negate tokens using the operator `!`. In Xtext, this allows recognizing the *complement* of a set of characters (or anything 'but' what is listed in the negation), very much akin to a negated character class in regular expressions. For example, the following terminal would recognize zero or more characters between two # symbols.
-```
-terminal BETWEEN_HASHES:'#' (!'#')* '#';
-```
+It is possible to negate tokens using the operator `!`. In Langium this produces a *negative lookahead*. I.e., it does *not* consume tokens, but it is a 'guard' for what the following expression can recognize.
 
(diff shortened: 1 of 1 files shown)


hugo/content/docs/grammar-language.md
```diff
@@ -478,14 +487,24 @@ terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
...
-It is possible to negate all tokens using the operator `!`, in regular expression the operator `^` is used for negation within a character class.
+It is possible to negate tokens using the operator `!`. In Xtext, this allows recognizing the *complement* of a set of characters (or anything 'but' what is listed in the negation), very much akin to a negated character class in regular expressions. For example, the following terminal would recognize zero or more characters between two # symbols.
 ```
 terminal BETWEEN_HASHES:'#' (!'#')* '#';
 ```
-In regular expression:
+
+This also works as expected when you use alternation in Xtext, `!('#'|'@')`, to recognize anything but `#` or `@`.
+
+*However, this works differently in Langium!* Instead of recognizing characters that are *not* in the set, it instead produces a negative lookahead. I.e., it does *not* consume tokens, but instead is a 'guard' for what the following expression can recognize. If you want to recognize a word that doesn't start with `no`, then you could write it in EBNF like so:
```
pluralia, line 497:
```suggestion
*However, this works differently in Langium!* Instead of recognizing characters that are *not* in the set, it produces a negative lookahead. I.e., it does *not* consume tokens, but it is a 'guard' for what the following expression can recognize. If you want to recognize a word that doesn't start with `no`, then you could write it in EBNF like so:
```

pluralia (commented):
I'd suggest explaining this topic to Langium first, and only then comparing it to Xtext: new users do not need to read about Xtext

montymxb:
Updated with review notes + reword

Issue #58, opened 2022-07-15
Create guide for scoping


msujew:
Scoping is arguably one of the most important, but nevertheless complex features of Langium. We should prioritize writing a guide for it.

PR #79, opened 2022-10-14
Add guides for scoping

msujew:
Closes https://github.com/langium/langium-website/issues/58

### DIFF
--- a/hugo/content/guides/scoping/_index.md
+++ b/hugo/content/guides/scoping/_index.md
@@ -16,24 +16,11 @@ y = 3; // Cannot link, `y` isn't in any of the available scopes
 ```
 
 By default, Langium creates a simplistic, purely block based scoping for your language as shown above. You can use it to reference all elements that are either declared in the current block or in one of its parents.
-However, this behavior can be easily modified to your domain needs, which is required for most languages. In general, resolving references is split into three distinct phases.
+However, this behavior can be easily modified to your domain needs, which is required for most languages.
 
-## Pre-processing
+In general, the reference resolving mechanism is split into three distinct phases of the document lifecycle. [Symbol indexing](/docs/document-lifecycle#symbol-indexing) is responsible for making referencable objects globally available. [Scope computation](/docs/document-lifecycle#computing-scopes) determines which elements are reachable at a certain position in a document. Finally, the [linking phase](/docs/document-lifecycle#linking) eagerly links each reference within a document to its target using scoping rules.
 
-Right after parsing a document, Langium starts a scope pre-processing phase. Before we are able to create references from one element of our language to another, we first have to define which elements are exposed to be referenced and under what name.
+In the following, we will look at different scoping kinds and styles and see how we can achieve them using Langium:
 
-This is the job of the `ScopeComputation` interface. It returns a map that points for each element of a given document to a list of node descriptions which are contained within that element.
-
-```ts
-namespace MyNamespace {
-    function A();
-    function B();
-    function C();
-}
-```
-
-Let's take the code snippet above as an example: The scope computation takes the `MyNamespace` element, creates descriptions (i.e. pointers to a node) for all child elements - functions A, B and C - and associates the `MyNamespace` element with its child descriptions.
-
-## Linking
-
-## Scoping
+1. [Qualified Name Scoping](./qualified-name)
+2. [Class Member Scoping](./class-member)
--- a/hugo/content/guides/scoping/_index.md
+++ b/hugo/content/guides/scoping/_index.md
@@ -1,0 +1,5 @@
+---
+title: "Scoping"
+weight: 100
+---
+
--- a/hugo/content/guides/scoping/_index.md
+++ b/hugo/content/guides/scoping/_index.md
@@ -3,3 +3,37 @@ title: "Scoping"
 weight: 100
 ---
 
+The process of scoping determines which elements can be referenced at a certain position of a document. You likely know scopes from programming, where some variables are only available from certain scopes:
+
+```ts
+let x = 42;
+x = 3; // References the `x` defined in the previous line
+
+if (something) {
+    let y = 42;
+}
+y = 3; // Cannot link, `y` isn't in any of the available scopes
+```
+
+By default, Langium creates a simplistic, purely block based scoping for your language as shown above. You can use it to reference all elements that are either declared in the current block or in one of its parents.
+However, this behavior can be easily modified to your domain needs, which is required for most languages. In general, resolving references is split into three distinct phases.
+
+## Pre-processing
+
+Right after parsing a document, Langium starts a scope pre-processing phase. Before we are able to create references from one element of our language to another, we first have to define which elements are exposed to be referenced and under what name.
+
+This is the job of the `ScopeComputation` interface. It returns a map that points for each element of a given document to a list of node descriptions which are contained within that element.
+
+```ts
+namespace MyNamespace {
+    function A();
+    function B();
+    function C();
+}
+```
+
+Let's take the code snippet above as an example: The scope computation takes the `MyNamespace` element, creates descriptions (i.e. pointers to a node) for all child elements - functions A, B and C - and associates the `MyNamespace` element with its child descriptions.
+
+## Linking
+
+## Scoping
--- a/hugo/content/guides/scoping/_index.md
+++ b/hugo/content/guides/scoping/_index.md
@@ -3,24 +3,29 @@ title: "Scoping"
 weight: 100
 ---
 
-The process of scoping determines which elements can be referenced at a certain position of a document. You likely know scopes from programming, where some variables are only available from certain scopes:
+You like know scopes from programming, where some variables are only available from certain areas (such as blocks) in your program. For example, take the short Typescript snippet below. Based on the block (scope) where a variable is declared, it may or may not be available at another location in the same program.
 
 ```ts
 let x = 42;
 x = 3; // References the `x` defined in the previous line
 
-if (something) {
+if (condition) {
     let y = 42;
 }
 y = 3; // Cannot link, `y` isn't in any of the available scopes
 ```
 
-By default, Langium creates a simplistic, purely block based scoping for your language as shown above. You can use it to reference all elements that are either declared in the current block or in one of its parents.
-However, this behavior can be easily modified to your domain needs, which is required for most languages.
+This kind of behavior is called lexical scoping. Although this default scoping implementation is suitable for prototyping -- and for some simple languages once finished -- this behavior can be easily modified to fit the needs of your language's domain.
 
-In general, the reference resolving mechanism is split into three distinct phases of the document lifecycle. [Symbol indexing](/docs/document-lifecycle#symbol-indexing) is responsible for making objects globally available for referencing. [Scope computation](/docs/document-lifecycle#computing-scopes) determines which elements are reachable at a certain position in a document. Finally, the [linking phase](/docs/document-lifecycle#linking) eagerly links each reference within a document to its target using scoping rules.
+In general, the way we resolve references is split into three phases of the document lifecycle:
+- [Symbol indexing](/docs/document-lifecycle#symbol-indexing) is responsible for making objects globally available for referencing.
+- [Scope computation](/docs/document-lifecycle#computing-scopes) determines which elements are reachable from a given position in your document.
+- Finally, the [linking phase](/docs/document-lifecycle#linking) eagerly links each reference within a document to its target using your language's scoping rules.
 
-In the following, we will look at different scoping kinds and styles and see how we can achieve them using Langium:
+In this guide, we'll look at different scoping kinds and styles and see how we can achieve them using Langium:
 
 1. [Qualified Name Scoping](./qualified-name)
 2. [Class Member Scoping](./class-member)
+
+Note that these are just example implementations for commonly used scoping methods.
+The scoping API of Langium is designed to be flexible and extensible for any kind of use case.
--- a/hugo/content/guides/scoping/_index.md
+++ b/hugo/content/guides/scoping/_index.md
@@ -3,7 +3,7 @@ title: "Scoping"
 weight: 100
 ---
 
-You like know scopes from programming, where some variables are only available from certain areas (such as blocks) in your program. For example, take the short Typescript snippet below. Based on the block (scope) where a variable is declared, it may or may not be available at another location in the same program.
+You likely know scopes from programming, where some variables are only available from certain areas (such as blocks) in your program. For example, take the short Typescript snippet below. Based on the block (scope) where a variable is declared, it may or may not be available at another location in the same program.
 
 ```ts
 let x = 42;
--- a/hugo/content/guides/scoping/_index.md
+++ b/hugo/content/guides/scoping/_index.md
@@ -18,7 +18,7 @@ y = 3; // Cannot link, `y` isn't in any of the available scopes
 By default, Langium creates a simplistic, purely block based scoping for your language as shown above. You can use it to reference all elements that are either declared in the current block or in one of its parents.
 However, this behavior can be easily modified to your domain needs, which is required for most languages.
 
-In general, the reference resolving mechanism is split into three distinct phases of the document lifecycle. [Symbol indexing](/docs/document-lifecycle#symbol-indexing) is responsible for making referencable objects globally available. [Scope computation](/docs/document-lifecycle#computing-scopes) determines which elements are reachable at a certain position in a document. Finally, the [linking phase](/docs/document-lifecycle#linking) eagerly links each reference within a document to its target using scoping rules.
+In general, the reference resolving mechanism is split into three distinct phases of the document lifecycle. [Symbol indexing](/docs/document-lifecycle#symbol-indexing) is responsible for making objects globally available for referencing. [Scope computation](/docs/document-lifecycle#computing-scopes) determines which elements are reachable at a certain position in a document. Finally, the [linking phase](/docs/document-lifecycle#linking) eagerly links each reference within a document to its target using scoping rules.
 
 In the following, we will look at different scoping kinds and styles and see how we can achieve them using Langium:
 
--- a/hugo/content/guides/scoping/class-member.md
+++ b/hugo/content/guides/scoping/class-member.md
@@ -1,0 +1,68 @@
+---
+title: "Class Member Scoping"
+weight: 200
+---
+
+For this guide, we will use excerpts from the [langium-lox](https://github.com/langium/langium-lox) project.
+It implements a strongly-typed version of the [Lox language](https://craftinginterpreters.com/the-lox-language.html) from the popular book [Crafting Interpreters](https://craftinginterpreters.com/).
+
+
+
+```ts
+type NamedElement = Parameter | FunctionDeclaration | VariableDeclaration | MethodMember | FieldMember | Class;
+
+MemberCall infers Expression:
+    // Reference a named element of our grammar
+    // Variables, functions, etc.
+    element=[NamedElement:ID]
+    // Create a new `MemberCall` and assign the old one to the `previous` property
+	({infer MemberCall.previous=current} 
+        // Member call with function call
+        ("." element=[NamedElement:ID] (
+		explicitOperationCall?='('
+		(
+		    arguments+=Expression (',' arguments+=Expression)*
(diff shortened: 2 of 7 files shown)


msujew:
@snarkipus do you mind me asking for a favor here? Can you read through the scoping guide (see [here for the PR deployment](https://deploy-preview-79--langium-website.netlify.app/guides/scoping/)) and give us feedback on how much that helps you understand scoping?

Lotes:
General question about `AstNodeDescriptions`: They have a name and a AST node (as URI address). You are using a full qualified name with '::' in-between. Which service is asking for full qualified names? When you have a reference like `Namespace::func`, I assumed that you result first `Namespace` and after that resolving `func` which is based on the search before.

Lotes:
1. Some visualization could be nice (or a table how the scopes/descriptions should look like).
2. This topic is really hard to grasp. Also for the `MemberCall`: You should only show one instance (even if both are the same situation) or maybe something simpler.
3. The scope computation saves scopes per AST node. Can you give an example? What makes sense here?
4. `ScopeComputation`, is it really needed? I think it is only timesaving, right? You only need scope provider and index, I believe...

Lotes:
5. Maybe 'out of scope' for this guide, but could you sketch what is called when and how often? `getScope` is called once for each cross-reference. But what about the index and the scope computation?

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
+---
+title: "Class Member Scoping"
+weight: 200
+---
+
+In this guide we will take at member based scoping. It's a mechanism you are likely familiar with from object oriented languages such as Java, C# and JavaScript:
```
montymxb, line 6:
```suggestion
In this guide we will take a look at member based scoping. It's a mechanism you are likely familiar with from object oriented languages such as Java, C# and JavaScript:
```

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
...
+
+void test() {
+    A a = new A();
+    B b = a.b; // Refers to the `b` defined in class `A`
+    String value = b.value; // Refers to the `value` defined in class `B`
+}
+```
+
+Member based scoping like this requires not only a modification of the default scoping provider, but also some other prerequisites.
+This includes a member call mechanism in your grammar and a rudamentary type system.
```
montymxb, line 25:
```suggestion
This includes adding a member call mechanism in your grammar and a rudimentary type system.
```

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
...
+For this guide, we will use excerpts from the [langium-lox](https://github.com/langium/langium-lox) project.
```
montymxb, line 26:
```suggestion
For this guide, we will use excerpts from the [langium-lox](https://github.com/langium/langium-lox) project to demonstrate how you can set this up yourself.
```

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
...
+It implements a strongly-typed version of the [Lox language](https://craftinginterpreters.com/the-lox-language.html) from the popular book [Crafting Interpreters](https://craftinginterpreters.com/).
```
montymxb, line 27:
```suggestion
This project implements a strongly-typed version of the [Lox language](https://craftinginterpreters.com/the-lox-language.html) from the popular book [Crafting Interpreters](https://craftinginterpreters.com/).
```

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
...
+
+We first start with the `MemberCall` grammar rule. It references one of our `NamedElements`, which could be variable declarations, functions, classes or methods and fields of those classes. Additionally, we allow function calls. Note that the grammar has no notion of whether these elements can actually be executed as functions. Instead, we always allow function calls on every named element and simply provide validation errors in case they were used erroneously. After parsing the first member call, we continue parsing further members as long as the input text provides us with further references to elements, separated by dots.
```
montymxb, line 29:
```suggestion
We'll first start with the `MemberCall` grammar rule, which references one of our `NamedElements`. These elements could be variable declarations, functions, classes or methods and fields of those classes. Additionally, we want to allow function calls on elements. Note that the grammar has no notion of whether these elements can actually be executed as functions. Instead, we always allow function calls on every named element, and simply provide validation errors in case an element is called erroneously. After parsing the first member call, we continue parsing further members as long as the input text provides us with further references to elements; which are separated by dots.
```

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
...
+
+```ts
+type NamedElement = FunctionDeclaration | VariableDeclaration | MethodMember | FieldMember | Class;
+
+MemberCall:
```
montymxb, line 34: This is a pretty dense parser rule. It might make sense to split it into 3 parts, introducing how each element of the rule works at a time, and then indicating that these are all part of the same `MemberCall` rule.

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
...
+        | (
+		explicitOperationCall?='('
+		(
+		    arguments+=Expression (',' arguments+=Expression)*
+		)?
+		')'))
+    )*;
+```
+
+A very important aspect of these chained member calls is the action which rewrites the resulting AST. In this case, it effectively reverses the direction of member call AST nodes. Instead of starting with the first encountered member call and then traversing down to the last, we start with the last and traverse the list of member calls up using the `previous` property. The reason for doing this becomes clear when looking at the scope provider for the Lox language:
```
montymxb, line 60:
```suggestion
A very important aspect of these chained member calls is the action which rewrites the resulting AST. In this case, it reverses the direction of member call AST nodes. Instead of starting with the first encountered member call and then traversing down to the last, we start with the last and traverse the list of member calls up using the `previous` property. The reason for doing this becomes clear when looking at the scope provider for the Lox language:
```

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
...
+    private scopeClassMembers(classItem: Class): Scope {
+        // Since Lox allows class-inheritance,
+        // we also need to look at all members of possible super classes for scoping
+        const allMembers = getClassChain(classItem).flatMap(e => e.members);
+        return this.createScopeForNodes(allMembers);
+    }
+}
+```
+
+When trying to compute the type of an expression, we are actually only interested in the final piece of the member call. However, to derive the type and scope of the final member call, we have to recursive identify the type of the previous member call. This is done by looking at the member call stored in the `previous` property and inferring its type. See [here](https://github.com/langium/langium-lox/blob/main/src/language-server/type-system/infer.ts) for the full implementation of the type inference system. Note that type inference requires scoping. This leads us to the following invocation graph:
```
montymxb, line 93:
```suggestion
When trying to compute the type of an expression, we are only interested in the final piece of the member call. However, to derive the type and scope of the final member call, we have to recursively identify the type of the previous member call. This is done by looking at the member call stored in the `previous` property and inferring its type. See [here for the full implementation of the type inference system in Lox](https://github.com/langium/langium-lox/blob/main/src/language-server/type-system/infer.ts). Note that type inference requires scoping.

With our scope provider in place, we can visualize how it interacts with Langium's implementation of Lox in the following invocation graph:
```

hugo/content/guides/scoping/class-member.md
```diff
@@ -0,0 +1,105 @@
...
+graph LR;
+A(<b>Lox Scope Provider</b><br><i>Member Call</i>) 
+--> B(<b>Type Inference</b><br><i>Class</i>)
+B --> C(<b>Lox Scope Provider</b><br><i>Member Call</i>)
+C --> D(<b>Type Inference</b><br><i>Class</i>)
+D --> E(...)
+E --> F(<b>Lox Scope Provider</b><br><i>Variable Declaration</i>)
+{{</mermaid>}}
+
+We recursively alternate between the scope provider and the type inference system until we arrive at a member call without any previous member call. At this point we resolve the reference using the default lexical scoping which is builtin into Langium. Note that it is recommended to cache type inference information as this naive approach to inference can quickly lead to performance issues.
```
montymxb, line 105: Might be also nice to show a quick example, just 2 member calls deep or something, and then connect it with this diagram. Can use one of the examples from above.

hugo/content/guides/scoping/qualified-name.md
```diff
@@ -0,0 +1,144 @@
...
+            // Iteratively prepend the name of the parent namespace
+            // This allows us to work with nested namespaces
+            name = `${parent.name}::${name}`;
+            parent = parent.$container;
+        }
+        return name;
+    }
+```
+
+Once we start exporting functions using their fully qualified name, references such as `QualifiedName::target` start working correctly. We can even nest multiple namespaces to create `Fully::Qualified::Name::target`. However, this leads us to another problem. We can now **only** reference functions using their fully qualified names, even if they're locally available:
```
montymxb, line 77:
```suggestion
Once we start exporting functions using their fully qualified name, references such as `QualifiedName::target` will start working correctly. We can even nest multiple namespaces to create `Fully::Qualified::Name::target`. However, this leads us to another problem. We can now only reference functions using their fully qualified names, even if they're locally available:
```

hugo/content/guides/scoping/qualified-name.md
```diff
@@ -0,0 +1,144 @@
...
+    void test() {
+        // Will not link correctly
+        target(); 
+        // Requires the new fully qualified name
+        QualifiedName::target();
+    }
+}
+```
+
+To rectify this change, we have to override the `computeLocalScopes` method. It serves as a way of providing access to elements which aren't exported globally. We can also use it to provide secondary access to the globally available objects using a local name.
```
montymxb, line 91:
```suggestion
To rectify this change, we have to override the `computeLocalScopes` method. It serves as a way of providing access to elements which aren't exported globally. We can also use it to provide secondary access to the globally available objects using a local name.
```
```suggestion
To rectify this problem, we have to override the `computeLocalScopes` method, which provides access to elements that aren't exported globally. We can also use this method to provide secondary access to globally available objects using a local name.
```

hugo/content/guides/scoping/qualified-name.md
```diff
@@ -0,0 +1,144 @@
...
+
+```ts
+export class CppScopeComputation extends DefaultScopeComputation {
+
+    // Emitting previous implementation for brevity
+
+    override async computeLocalScopes(document: LangiumDocument): Promise<PrecomputedScopes> {
+        const model = document.parseResult.value as CppProgram;
+        // This map stores a list of descriptions for each node in our document
```
montymxb, line 100:
```suggestion
        // This multi-map stores a list of descriptions for each node in our document
```

hugo/content/guides/scoping/qualified-name.md
```diff
@@ -0,0 +1,144 @@
...
+        for (const element of container.elements) {
+            if (isFunctionDeclaration(element)) {
+                // Create a simple local name for the function
+                const description = this.descriptions.createDescription(element, element.name, document);
+                localDescriptions.push(description);
+            } else if (isNamespace(element)) {
+                const nestedDescriptions = this.processContainer(element, scopes, document, cancelToken);
+                for (const description of nestedDescriptions) {
+                    // Add qualified names to the container
+                    // This could also be a partial qualified name
```
montymxb, line 121:
```suggestion
                    // This could also be a partially qualified name
```

hugo/content/guides/scoping/qualified-name.md
```diff
@@ -0,0 +1,144 @@
...
+        document: LangiumDocument
+    ): AstNodeDescription {
+        // `getQualifiedName` has been implemented in the previous section
+        const name = this.getQualifiedName(container.name, description.name);
+        return this.descriptions.createDescription(description.node!, name, document);
+    }
+}
+```
+
+This new change now allows us to use local names of functions in the local scope, while they are still being exported using their fully qualified name to the global scope.
```
montymxb, line 143:
```suggestion
This new change now allows us to use local names of functions in the local scope, while they are still exported using their fully qualified name to the global scope.
```

montymxb (commented):
Some initial points. In terms of the suggested changes, you don't have to add them in verbatim, they're more of an easy to highlight typos or issues in larger blocks of text.

hugo/content/guides/scoping/class-member.md
montymxb: Also, geekdoc has [admonitions](https://geekdocs.de/asciidoc/admonitions/) for cases like `NOTE`, `TIP`, `IMPORTANT`, etc. These could be nice to add in places where you use `Note` as is.

snarkipus:
@msujew I apologize for the late response - somehow the original tag slipped through the inbox. I'd be happy to take a look at it and provide feedback/contribute. Consequently, I have a fair number of simple diagrams (most likely not 100% correct) which may be helpful (e.g. [document life cycle](https://app.excalidraw.com/l/7YrgtG5Ew9f/zXKm1VyNUa), [Default Scoping & Indexing](https://app.excalidraw.com/l/7YrgtG5Ew9f/8FbIl3t3Cnl)).

snarkipus:
@msujew : I'll split my thoughts into two comments: one which addresses some of the comments and a separate comment which summarizes what I think would be/could be helpful.

**RE: @montymxb  comments:**

> 1.  Some visualization could be nice (or a table how the scopes/descriptions should look like).

 - 100% agree with this - I linked a couple sketches that I did in my previous comment. Here's an example of another one from when I was working through this exact topic ([LINK](https://app.excalidraw.com/l/7YrgtG5Ew9f/3b9cq0wMqlE)). Note, these are quite obviously working diagrams and almost guaranteed not to be correct.

> 2.  This topic is really hard to grasp. Also for the `MemberCall`: You should only show one instance (even if both are the same situation) or maybe something simpler.

 - From my personal experience, this particular topic is _incredibly difficult_ and should absolutely be treated as an _advanced topic_.

> 3.  The scope computation saves scopes per AST node. Can you give an example? What makes sense here?

 - I think one of the diagrams I linked illustrates this - if not, one could be easily created.

> 4.  `ScopeComputation`, is it really needed? I think it is only timesaving, right? You only need scope provider and index, I believe...

 - This could go either way, it really all depends on how you would like to explain it.

> 5. Maybe 'out of scope' for this guide, but could you sketch what is called when and how often? `getScope` is called once for each cross-reference. But what about the index and the scope computation?

 - See linked diagrams.

In an effort to avoid further cluttering up this PR, can we open a separate discussion thread?

hugo/content/guides/scoping/qualified-name.md
```diff
@@ -0,0 +1,144 @@
...
+// Scope computation for C++
+export class CppScopeComputation extends DefaultScopeComputation {
+
+    constructor(services: LangiumServices) {
+        super(services);
+    }
+}
+
+// Services module for overriding the scope computation
+export const CppModule: Module<CppServices, PartialLangiumServices & CppAddedServices> = {
```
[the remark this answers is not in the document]
montymxb: Non-blocking, marking as resolved

hugo/content/docs/document-lifecycle.md
```diff
@@ -49,12 +49,12 @@ B --> C(<b>LangiumParser</b><br><br><i>parses a string into an AST</i>)
...
+Indexing of the exported symbols of an AST is executed on documents with the state `Parsed`. The default `AstNodeDescriptionProvider` service creates an `AstNodeDescription` for the root node (i.e. the node created by parsing the entry rule) and each named `AstNode` directly descending from the root node. This `AstNodeDescription` contains the `type` of the node, its identifier (i.e. the `name` property), the URI of the document where the node is located, and the location of the node inside of the document. The generated set of `AstNodeDescription`s makes symbols from a `LangiumDocument` accessible to other documents in the same workspace.
```
montymxb, line 55: Should the last sentence be revised to refer to the single AstNodeDescription rather than the set?

hugo/content/guides/scoping/_index.md
```diff
@@ -3,24 +3,29 @@ title: "Scoping"
 weight: 100
 ---
 
-The process of scoping determines which elements can be referenced at a certain position of a document. You likely know scopes from programming, where some variables are only available from certain scopes:
+You like know scopes from programming, where some variables are only available from certain areas (such as blocks) in your program. For example, take the short Typescript snippet below. Based on the block (scope) where a variable is declared, it may or may not be available at another location in the same program.
```
montymxb, line 6: Typo on 'like'
```suggestion
You likely know scopes from programming, where some variables are only available from certain areas (such as blocks) in your program. For example, take the short Typescript snippet below. Based on the block (scope) where a variable is declared, it may or may not be available at another location in the same program.
```

hugo/content/guides/scoping/class-member.md
```diff
@@ -35,19 +35,25 @@ MemberCall:
     // Reference a named element of our grammar
     // Variables, functions, etc.
     element=[NamedElement:ID]
+    // Parse an operation call on this element
     (explicitOperationCall?='(' (
+        // Parse any arguments for the operation call
```
montymxb, line 40: indentation here

hugo/content/guides/scoping/qualified-name.md
```diff
@@ -3,24 +3,26 @@ title: "Qualified Name Scoping"
...
-    // Should call the `target` function defined in the `QualifiedName` namespace
-    QualifiedName::target();
+void main() {
+    // Should call the `getDocumentation` function defined in the `Langium` namespace
+    Langium::getDocumentation();
 }
 ```
 
-This behavior can be achieved in Langium by exporting the `target` function under the name `QualifiedName::target`. We will first set up a new `ScopeComputation` class that extends the `DefaultScopeComputation` and bind it in our module:
+As can be seen, using qualified name scoping is quite helpful in this case. It allows us to reference the `getDocumentation` function through the scope computed & made available by the `Langium` namespace, even though it's not directly accessible within the scope of `main` by itself. This behavior can be achieved in Langium by exporting the `getDocumentation` function under the name `Langium::getDocumentation`. To do this, we will first set up a new `ScopeComputation` class that extends the `DefaultScopeComputation`, and will be responsible for our custom scope computation. Then, we'll want to bind our custom scope computation class in our module:
```
montymxb, line 22: Small rewrite suggestion for 2nd to last sentence
```suggestion
As can be seen, using qualified name scoping is quite helpful in this case. It allows us to reference the `getDocumentation` function through the scope computed & made available by the `Langium` namespace, even though it's not directly accessible within the scope of `main` by itself. This behavior can be achieved in Langium by exporting the `getDocumentation` function under the name `Langium::getDocumentation`. To do this, we will first set up a new `ScopeComputation` class that extends the `DefaultScopeComputation`. This class will be responsible for our custom scope computation. Then, we'll want to bind our custom scope computation class in our module:
```

montymxb (commented):
Looks really good! Just a handful of things at this point.

Not blocking, but it might also be nice to color parts of the sequence diagram, like `type` and `node` as an example.

hugo/content/guides/scoping/qualified-name.md
```diff
@@ -0,0 +1,230 @@
...
+     * Export all functions using their fully qualified name
+     */
+    override async computeExports(document: LangiumDocument): Promise<AstNodeDescription[]> {
+        const exportedDescriptions: AstNodeDescription[] = [];
+        for (const childNode of streamAllContents(document.parseResult.value)) {
+            if (isFunctionDeclaration(childNode)) {
+                const fullyQualifiedName = this.getQualifiedName(childNode, childNode.name);
+                // `descriptions` is our `AstNodeDescriptionProvider` defined in `DefaultScopeComputation`
+                // It allows us to easily create descriptions that point to elements using a name.
+                exportedDescriptions.push(this.descriptions.createDescription(modelNode, fullyQualifiedName, document));
```
pluralia, line 59:
```suggestion
             exportedDescriptions.push(this.descriptions.createDescription(childNode, fullyQualifiedName, document));
```

pluralia:
Currently, 2 examples are used to explain the scoping of qualified names: a custom example is used first, followed by a reference to the domain model example. I'd suggest using only the domain model example to focus on one example -- it's easier for understanding.

spoenemann (approved):
Let's ship this so it's ready for the v1.0 release.

PR #80, opened 2022-10-15
Fix comment order

rhumbertgz:
The comment "// should be replaced by:" was pointing to the generated JS code instead of the refactored code. I just added the right comment for the generated JS code, and I moved the original comment to the right position


msujew (commented):
See my comment in https://github.com/langium/langium-website/pull/81. The documentation is correct as it is right now.

rhumbertgz:
Thanks for your explanation 👍

PR #81, opened 2022-10-15
Move generated code after dummy rules' declaration

rhumbertgz:
Right now the generated code is in another listing

### DIFF
--- a/hugo/content/docs/ast-types.md
+++ b/hugo/content/docs/ast-types.md
@@ -329,15 +329,16 @@ Let's look at two dummy rules:
 X : A | B;
 
 Y: name=ID;
-```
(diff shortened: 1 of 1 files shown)


msujew (commented):
Thanks for the contribution, though there might be a misunderstanding on your part. The code in question isn't generated code but rather source code in the grammar file. Declaring the types explicitly allows us to remove dummy rules, which are introduced purely for simplicity of cross references.

You can write the following in your grammar file:

```
type X = A | B;

interface Y {
    name: string
}

Y returns Y: name=ID;
```

which then generates this code for the `ast.ts`:

```
type X = A | B;

interface Y {
    name: string
}
```

rhumbertgz:
Thanks for your explanation 👍

PR #82, opened 2022-10-16
Improve dummy rule documentation

msujew:
The current documentation explains fairly well how to refactor dummy rules, but doesn't really explain why they are sometimes needed in the first place.

This change addresses this, as it proposes a use case, where we first introduce a problem, which can be solved by using a dummy rule. Afterwards, we refactor our solution using declared types to remove the dummy rule again.

### DIFF
--- a/hugo/content/docs/sematic-model.md
+++ b/hugo/content/docs/sematic-model.md
@@ -187,6 +187,7 @@ Primary infers Expression:
     '(' Expression ')' | {Literal} value=NUMBER;
 ```
 When parsing a document containing `def x: (1 + 2) + 3`, this is the shape of the semantic model node:
+
 {{<mermaid>}}
 graph TD;
 expr((expr)) --> left((left))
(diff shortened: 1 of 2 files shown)


pluralia (commented):
Looks good! I'd propose:
* leave explicit definition of the "dummy rules"
* the source doc contains an example of eliminating different types of dummy rules (to type and to interface) -- I think it's nice to add a second example in the edited doc, where a dummy rule is converted to an interface

pluralia:
> I agree, I just call the section Reference Unions now, as that accurately describes what the section is trying to accomplish.

Do you think to eliminate the term "dummy rules" completely?

hugo/content/docs/sematic-model.md
```diff
@@ -319,25 +327,44 @@ X:
...
+
+We can also use interfaces in place of union types with similar results:
+
+```
+interface NamedElement {
     name: string
 }
 
-Y returns Y: name=ID;
+// Infers an interface `Function` that extends `NamedElement`
```
pluralia, line 365: I'd suggest to tell about the difference in using `NamedElement` as an interface and as a type. Specifically, that different types will be generated and there will be no `Function` type if you use the syntax `Function returns NamedElement`.

pluralia (commented):
I like your examples :)

spoenemann (approved):
Let's ship this so it's ready for the v1.0 release.

PR #83, opened 2022-10-17
Make playground visible for mobile devices

Lotes:


Lotes:
I reused an old branch. Maybe that is why the history is so long O.o

hugo/layouts/partials/langium-mobile-menu.html
```diff
@@ -3,21 +3,18 @@
 x-transition:enter-start="opacity-0 scale-95" x-transition:enter-end="opacity-100 scale-100"
 x-transition:leave="transition ease-in duration-100 transform"
 x-transition:leave-start="opacity-100 scale-100" x-transition:leave-end="opacity-0 scale-95"
-class="absolute top-0 inset-x-0 p-2 transform origin-top-right md:hidden">
+class="absolute inset-x-0 top-0 p-2 origin-top-right transform md:hidden">
```
montymxb, line 6: Is there a particular reason why the css classes have been reordered?

montymxb (commented):
Question about the css classes, beyond that the single navbar change looks good. If the classes don't need to be switched around, it would be nice to remove the extra changes in this PR. Although if they're needed, we can add them in too.

hugo/layouts/partials/langium-mobile-menu.html
Lotes: Good point! No it happens automatically on save...

Lotes: I disabled my Headwind extension. It was once required for another customer.

montymxb (approved):
Approved 🧙‍♂️

PR #84, opened 2022-10-20
Move monaco-workers to libs directory below public

kaisalmen:
Both showcase and playground use the monaco-editor-workers now copied to **public/libs**. Only the single editor `editorWorker-iife.js` is now copied.

Removed unused terminal rules as well.


msujew (approved):
LGTM 👍

montymxb:
I'll bring this in

Issue #89, opened 2022-10-28
Playground: Add "Share" button


spoenemann:
Add a "Share" button that generates a URL with parameters for the left and right editor contents. Also add a "Copy" button to copy the generated URL to the clipboard.

Example URL: `https://langium.org/playground/?grammarContent=ABCD&targetLangContent=EFGH`

spoenemann:
I once made something similar for the ELK demonstrator using LZString to compress the editor content:

https://github.com/kieler/elk-live/blob/d96c5e0aa872c528098cb9d57c812ce46fef9a21/client/src/elkgraph/editor.ts#L28

PR #91, opened 2022-10-29
Add share feature to playground

msujew:
Closes https://github.com/langium/langium-website/issues/89

Adds a button to the playground that puts a playground URL with compressed grammar and content into the clipboard. Any user who opens that URL will be presented with the original grammar and content.


spoenemann (commented):
Thanks, works great! The positioning of the Share button in the top bar was a bit surprising to me. Is there a way to put it below the editors?

Mobile mode looks broken:
Screenshot 2022-10-31 at 10 14 52

spoenemann (approved):
Looks good, thanks!

PR #92, opened 2022-11-03
Add Command Invocation Example via JSON-RPC

montymxb:
This adds a simple example of how a command can be invoked via JSON-RPC. To do so we also needed to updated the `monaco-editor-wrapper` package to new version, which exposes the `monaco-vscode-api` object (fittingly called `vscode` in this case).

Please note that the current command does not successfully execute! There has to be a handler registered on the server as well for the command to be received successfully, and a response to be generated. This serves more as a template for further work, particularly where this kind of logic can be moved into React, so it can handle when commands are executed, and handle the responses as well.


package.json
```diff
@@ -36,7 +36,7 @@
     "@babel/core": "^7.19.3",
     "@babel/preset-react": "^7.18.6",
     "@monaco-editor/react": "^4.4.6",
-    "monaco-editor-wrapper": "^1.2.0",
+    "monaco-editor-wrapper": "^1.3.1-next.0",
```
kaisalmen, line 39: We should switch to final 1.3.1 before merging. Also, if possible the whole `langium-preview-develop` branch should be rebased to the latest `main`.

kaisalmen (changes requested):
See comment, otherwise 👍

package.json
montymxb: Since this is still preliminary development, I think going with next is okay for *now*, and then we can include the later **1.3.1** (so long as there's no major breaking changes). Unless that's being released in the next day or two, then we can incorporate it before merging here

However It would be best to get something added first, and then update, as it's currently blocking further work.

kaisalmen: 1.3.1 is now available, so this won't block you if that is ok? 🙂

montymxb:
This PR is also pulling in commits from 'main' as a rebase. Syncs things up with changes there.

montymxb:
Closing this, as it's been superseded by the new monaco react component that we're putting together. That will be doing the same thing, but much easier to integrate with React directly.

PR #101, opened 2022-11-23
Add formatting guide

msujew:
Explains how an adopter would customize their own formatter for their language.

### DIFF
--- a/hugo/content/guides/formatting.md
+++ b/hugo/content/guides/formatting.md
@@ -1,0 +1,103 @@
+---
+title: "Formatting"
+weight: 300
+---
+
+Langium's formatting API allows to easily create formatters for your language.
+We start building a custom formatter for our language by creating a new class that inherits from `AbstractFormatter`.
+
+```ts
+import { AbstractFormatter, AstNode, Formatting } from 'langium';
+
+export class CustomModelFormatter extends AbstractFormatter {
+    protected format(node: AstNode): void {
+        // This method is called for every AstNode in a document
+    }
+}
+...
+// Bind the class in your module
+export const CustomModelModule: Module<CustomModelServices, PartialLangiumServices> = {
+    lsp: {
+        Formatter: () => new CustomModelFormatter()
+    }
+};
+```
+
+The entry point for the formatter is the abstract `format(AstNode)` method. The `AbstractFormatter` calls this method for every node of our model.
+To perform custom formatting for every type of node, we will use pattern matching.
+In the following example, we will take a closer look at a formatter for the [domain-model](https://github.com/langium/langium/tree/main/examples/domainmodel) language.
+In particular, we will see how we can format the root of our model (`DomainModel`) and each nested element (`Entity` and `PackageDeclaration`).
+
+To format each node, we use the `getNodeFormatter` method of the `AbstractFormatter`. The resulting generic `NodeFormatter<T extends AstNode>` provides us with methods to select specific parts of a parsed `AstNode` such as properties or keywords.
+
+Once we have selected the nodes of our document that we are interested in formatting, we can start applying a specific formatting.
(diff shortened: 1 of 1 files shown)


hugo/content/guides/formatting.md
```diff
@@ -0,0 +1,103 @@
+---
+title: "Formatting"
+weight: 300
+---
+
+Langium's formatting API allows to easily create formatters for your language.
+We start building a custom formatter for our language by creating a new class that inherits from `AbstractFormatter`.
+
+```ts
```
pluralia, line 9: Could you specify where to put the `CustomModelFormatter` class? Now it looks like it has to be in one file with `CustomModelModule`.

hugo/content/guides/formatting.md
```diff
@@ -0,0 +1,103 @@
...
+import { AbstractFormatter, AstNode, Formatting } from 'langium';
+
+export class CustomModelFormatter extends AbstractFormatter {
+    protected format(node: AstNode): void {
+        // This method is called for every AstNode in a document
+    }
+}
+...
+// Bind the class in your module
+export const CustomModelModule: Module<CustomModelServices, PartialLangiumServices> = {
```
pluralia, line 19: Coma in this line is highlighted:
Screenshot 2022-11-23 at 15 34 46

hugo/content/guides/formatting.md
```diff
@@ -0,0 +1,103 @@
...
+    // Prepend all these nodes with no indent
+    nodes.prepend(Formatting.noIndent());
+}
+```
+
+Our other elements, namely `Entity` and `PackageDeclaration`, can be arbitrarily deeply nested, so using `noIndent` is out of the question for them.
+Instead we will use `indent` on everything between the `{` and `}` tokens. The formatter internally keeps track of the current indentation level:
+
+```ts
+if (ast.isEntity(node) || isPackageDeclaration(node)) {
```
pluralia, line 57: This code differs with what we have for the domain model in the repo [here](https://github.com/langium/langium/blob/main/examples/domainmodel/src/language-server/domain-model-formatter.ts). Maybe update repo's version?

hugo/content/guides/formatting.md
```diff
@@ -0,0 +1,103 @@
...
+    formatter.interior(bracesOpen, bracesClose).prepend  (Formatting.indent());
+    // Also move the newline to a closing brace
+    bracesClose.prepend(Formatting.newLine());
+    // Surround the name property of the element
+    // With one space to each side
+    formatter.property("name").surround(Formatting.oneSpace());
+}
+```
+
+Note that most predefined `Formatting` methods accept additional arguments which make the resulting formatting more lenient.
```
pluralia, line 73: Do you mind to tell more about these methods? Yes, it's a guide and not a documentation, but maybe it's a nice to see a couple more examples of API usage.

hugo/content/guides/formatting.md
pluralia: I'd propose don't use `CustomModelFormatter` -- you do all explanations using the domain model, so, maybe it's nice to tell about `DomainModelFormatter` from the beggning.

pluralia: Also, looks like the `CustomModelFormatter` and `CustomModelModule` has to be called `CustomFormatter` and `CustomModule`.

hugo/content/guides/formatting.md
montymxb: This is an existing issue with the syntax highlighting. I've seen it in other changes as well. We should address it separately.

hugo/content/guides/formatting.md
msujew: We should do that, but it's unrelated to this PR.

pluralia (approved):
Approved ✅

PR #106, opened 2022-11-25
Updated monaco-editor-wrapper 1.3.2 Removing codicon.ttf loading error

kaisalmen:
codicon.ttf is now directly contained in the css encoded as base64 data url. Therefore only one helper function is needed and the path error in the console is gone.


msujew (approved):
I don't seem to get the path error, but the change looks good to me anyway 👍

PR #108, opened 2022-11-28
Make playground's layout more responsive

Lotes:
* using a [grid](https://play.tailwindcss.com/iynmIDdW0m) layout
* NOT using [flex-box](https://play.tailwindcss.com/4js449QH1r) because I want the neighbor cells having same measures

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,3 +7,5 @@ Our setup uses TailwindCSS to build the styles which are then copied into the Hu
 Please look into the sub folders [hugo/](hugo/README.md) and [tailwind/](tailwind/README.md).
 
 You can find the `showcase/` folder [here](hugo/content/showcase).
+
+And the `playground/` folder [here](hugo/content/playground).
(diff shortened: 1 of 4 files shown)


danieldietrich (commented):
Hi Lotes,

thank you for giving me the opportunity to review :)

I tested your changes on mobile and the responsive layout works really well from the UX perspective, both in portrait and landscape direction.

There is one question I asked myself: are there any (self-imposed) best practices regarding mixing inline TailwindCSS utility-classes and CSS style definitions? Personally, I would remove all `id`s from the HTML tags and add inline tailwind style instead where possible. The goal should be to have no CSS at all but only inline tailwind utility classes.

_(Disclaimer: I know that CSS and tailwind was mixed before this PR but maybe it is a good opportunity to think about that now)_

What do you think?

Thx & great work!
Daniel

Lotes:
I converted almost all styles to TailwindCSS. For making manipulations on the header and footer I need to do it with CSS, because they are global partials used by all sub pages. So I just injected the id for a footer and for the header and referenced these ids in my media query.

I would be happy about an approval, the next PR is in den Startlöchern.

danieldietrich (approved):
Looks super-gut to me!

PR #110, opened 2022-12-03
Sort package json content

kaisalmen:
Because the main changes were already merged by Emil, this PR only adds the `package.json` clean-up and that the latest version of `@typefox/monaco-editor-react/bundle` directly exposes `monaco`

I only declared `devDependencies` everywhere now, because all deps are only needed to create an output and therefore this seemed reasonable.


spoenemann (approved):
:+1:

emilkrebs:
Looks all good!

PR #111, opened 2022-12-07
Remove error color

msujew:
When Chroma encountered a TypeScript syntax error (even though there wasn't), it displayed a white box. This change removes the background color for errors.

### DIFF
--- a/hugo/static/custom.css
+++ b/hugo/static/custom.css
@@ -169,6 +169,9 @@
     }
 }
 
+:root .chroma .err {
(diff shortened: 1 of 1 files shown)


montymxb (commented):
This seems like it only catches the error in dark mode, but not light mode. Between the two, the former was much much harder to read, so this may be okay as is. Up to you about handling the lesser light mode case.
e1
e2

spoenemann (approved):

PR #117, opened 2022-12-15
Update customizing_cli.md

steve-hickman-epistimis:
path to example logo file fixed

### DIFF
--- a/hugo/content/tutorials/customizing_cli.md
+++ b/hugo/content/tutorials/customizing_cli.md
@@ -109,8 +109,13 @@ test()
 We'll save this under our project root as **test.logo**, and we can test that it's correct using our CLI like so:
 
 ```bash
+./bin/cli parseAndValidate test.logo
+```
+NOTE: The langium-minilogo repo places `test.logo` in an `examples` subdirectory. So, for that case, the CLI usage would be:
+```bash
 ./bin/cli parseAndValidate examples/test.logo
 ```
+It does not matter where you place your .logo files. Organize them as you see fit.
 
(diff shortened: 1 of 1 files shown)


msujew (commented):
The paragraph before that is kind of ambiguous in regards to what the "project root" is, as `example` might be the project root for the extension development process. @montymxb care to clarify?

montymxb:
Ah I see where this is coming from... The [examples](https://github.com/langium/langium-minilogo/tree/main/examples) folder is where **test.logo** and the other example programs are stored in langium-minilogo, but in the documentation here it's implied the example is created at the top level of the project (@msujew project root can be considered where the top-level package.json lives in this context, but maybe we need to clarify that).

This sounds like an okay change in terms of consistency with the repo, but we should also update the mention above of where we create this example program. I'll leave a note about that.

hugo/content/tutorials/customizing_cli.md
```diff
@@ -109,7 +109,7 @@ test()
 We'll save this under our project root as **test.logo**, and we can test that it's correct using our CLI like so:
```
montymxb, line 109: @steve-hickman-epistimis could you update this sentence to indicate that the program is saved under the examples folder instead? It wouldn't also be bad to point out that it's okay to save it elsewhere, so long as you keep in mind where it is 😉 .

montymxb (changes requested):
One small addition would be needed, to make sure the code snippet matches the description above. But the change makes this match up with the langium-minilogo repo, which is much appreciated 😄!

montymxb (approved):
LGTM :octocat:

PR #119, opened 2022-12-29
Update writing_a_grammar.md

insafuhrmann:
Corrected typos


montymxb (approved):
Super! I just was looking at some grammatical errors myself but overlooked these PRs. We can take these ones first.

PR #121, opened 2022-12-29
Update customizing_cli.md

insafuhrmann:
Typo fixes and a small wording change (as imho the sentence does not fully fit the example (no ast extraction).

### DIFF
--- a/hugo/content/tutorials/customizing_cli.md
+++ b/hugo/content/tutorials/customizing_cli.md
@@ -73,7 +73,7 @@ export const parseAndValidate = async (fileName: string): Promise<void> => {
 };
 ```
 
-Some amount of the contents for our custom action are shared with the `generateAction` function. This isn't surprising, given that for example we still need to set up our language's services.
+Some amount of the contents for our custom action are shared with the `generateAction` function. This isn't surprising given that we still need to set up our language's services.
 
 ## Building and Running the CLI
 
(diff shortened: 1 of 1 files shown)


hugo/content/tutorials/customizing_cli.md
```diff
@@ -73,7 +73,7 @@ export const parseAndValidate = async (fileName: string): Promise<void> => {
 };
 ```
 
-A good amount of the contents for our custom action are shared with the `generateAction` function. This isn't surprising, given that we still need to set up our language's services, and we still need to be able to extract the AST.
+Some amount of the contents for our custom action are shared with the `generateAction` function. This isn't surprising, given that for example we still need to set up our language's services.
```
montymxb, line 76: I'm on board with dropping the note about extracting the AST. However I would also suggest another slight change to the 2nd sentence, dropping for example and going for a simpler sentence.
```suggestion
Some amount of the contents for our custom action are shared with the `generateAction` function. This isn't surprising given that we still need to set up our language's services.
```

montymxb (changes requested):
Typo fixes look good, but small suggestion on one sentence.

insafuhrmann:
@montymxb 
Thanks for the review, I committed the requested changes.

montymxb (approved):
:octocat: approved!

insafuhrmann:
Thanks @montymxb!

PR #122, opened 2022-12-29
Update generation.md

insafuhrmann:
Fixed typos and slight changes to the examples (adjusted to signatures after trial run).


kaisalmen (approved):
This looks good. Just checked: state (`DrawingState`) is only used in the mini-logo repository.

montymxb:
Yep, the drawing state is vestigial parameter that should be factored out of the repo as well. After several versions there are still leftovers from the initial implementation floating around.

montymxb (approved):

PR #125, opened 2023-01-27
Use antlr syntax highlighting for langium code

msujew:
The antlr syntax highlighting is fairly close to Langium. We could also add some special syntax highlighting for Langium later, but this is a quick win improve readability of Langium code snippets.


spoenemann (approved):
:+1:

PR #126, opened 2023-02-02
Adds AST Deserializer + Example Usage

montymxb:
Updates the Statemachine react code to include a rough approach for de-serializing a Langium-based AST. In addition, this also includes some examples of how that AST can then be traversed to retrieve specific information, which could then drive the generation of UI elements as desired.


emilkrebs:
Looks perfect to me, I will merge it so I can continue working.

PR #128, opened 2023-02-05
fix documentation in langium_and_monaco.md

doemming:
fix shell command separation in npm build configuration snippet ("npx shx ...")


montymxb (approved):
Cool looks good 👍 . Thanks for the contribution!

While you're at it, would you also be willing to open up a similar PR for the [langium-minilogo](https://github.com/langium/langium-minilogo) repo? The package.json still reflects the old form, and it would be good to keep these changes in sync.

Issue #70, opened 2022-09-02
Create a guide on how to customize lexing


msujew:
Should contain a few words on

* Chevrotain lexing (first fit, etc)
* How to create a multi-mode lexer
* How to create tokens with custom lexing patterns

PR #132, opened 2023-02-13
Guide on multi-mode lexing

msujew:
Effectively closes https://github.com/langium/langium-website/issues/70

### DIFF
--- a/hugo/content/guides/multi-mode-lexing.md
+++ b/hugo/content/guides/multi-mode-lexing.md
@@ -7,51 +7,81 @@ Many modern programming languages such as [JavaScript](https://developer.mozilla
 They are a way to easily concatenate or interpolate string values while maintaining great code readability.
 This guide will show you how to support template literals in Langium.
 
+For this specific example, our template literal starts and ends using backticks `` ` ``  and are interupted by expressions that are wrapped in curly braces `{}`.
+So in our example, usage of template literals might look something like this:
+
+```js
+println(`hello {name}!`);
+```
+
+Conceptually, template strings work by reading a start terminal which starts with `` ` `` and ends with `{`, 
+followed by an expression and then an end terminal which is effectively just the start terminal in reverse using `}` and `` ` ``.
+Since we don't want to restrict users to only a single expression in their template literals, we also need a "middle" terminal reading from `}` to `{`.
+Of course, there's also the option that a user only uses a template literal without any expressions in there.
+So we additionally need a "full" terminal that reads from the start of the literal all the way to the end in one go.
+
+To achieve this, we will define a `TemplateLiteral` parser rule and a few terminals.
+These terminals will adhere to the requirements that we just defined.
+To make it a bit easier to read and maintain, we also define a special terminal fragment that we can reuse in all our terminal definitions:
+
 ```antlr
 TemplateLiteral:
     // Either just the full content
-    content+=TemplateContent | 
-    // Or template string parts with expressions in between
+    content+=TEMPLATE_LITERAL_FULL |
+    // Or template literal parts with expressions in between
     (
-        content+=TemplateContentStart 
-        content+=Expression? 
+        content+=TEMPLATE_LITERAL_START 
+        content+=Expression?
         (
-            content+=TemplateContentMiddle 
+            content+=TEMPLATE_LITERAL_MIDDLE
             content+=Expression?
-        )* 
-        content+=TemplateContentEnd
-    );
+        )*
+        content+=TEMPLATE_LITERAL_END
+    )
+;
 
-TemplateContent returns TextLiteral:
-    value=RICH_TEXT;
+terminal TEMPLATE_LITERAL_FULL:
+    '`' IN_TEMPLATE_LITERAL* '`';
 
-TemplateContentStart returns TextLiteral:
-    value=RICH_TEXT_START;
+terminal TEMPLATE_LITERAL_START:
+    '`' IN_TEMPLATE_LITERAL* '{';
 
-TemplateContentMiddle returns TextLiteral:
-    value=RICH_TEXT_INBETWEEN;
+terminal TEMPLATE_LITERAL_MIDDLE:
+    '}' IN_TEMPLATE_LITERAL* '{';
 
-TemplateContentEnd returns TextLiteral:
-    value=RICH_TEXT_END;
+terminal TEMPLATE_LITERAL_END:
+    '}' IN_TEMPLATE_LITERAL* '`';
 
-terminal RICH_TEXT:
-    '`' IN_RICH_TEXT* '`';
-
-terminal RICH_TEXT_START:
-    '`' IN_RICH_TEXT* '{';
-
-terminal RICH_TEXT_INBETWEEN:
-    '}' IN_RICH_TEXT* '{';
-
-terminal RICH_TEXT_END:
-    '}' IN_RICH_TEXT* '`';
-
-terminal fragment IN_RICH_TEXT:
+// '{{' is handled in a special way so we can escape normal '{' characters
+// '``' is doing the same for the '`' character
+terminal fragment IN_TEMPLATE_LITERAL:
     /[^{`]|{{|``/;
 ```
 
+If we go ahead and start parsing files with these changes, most things should work as expected.
+However, depending on the structure of your existing grammar, some of these new terminals might be in conflict with existing terminals of your language.
+For example, if your language supports block statements, chaining multiple blocks together will make this issue apparent:
+
+```js
+{
+    console.log('hi');
+}
+{
+    console.log('hello');
+}
+```
+
+The `} ... {` block in this example won't be parsed as separate `}` and `{` tokens, but instead as a single `TEMPLATE_LITERAL_MIDDLE` token, resulting in a parser error due to the unexpected token.
+This doesn't make a lot of sense, since we aren't in the middle of a template literal at this point anyway.
+However, our lexer doesn't know yet that the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals are only allowed to show up within a `TemplateLiteral` rule.
+To rectify this, we will need to make use of lexer modes. They will give us the necessary context to know whether we're inside a template literal or outside of it.
+Depending on the current selected mode, we can lex different terminals. In our case, we want to exclude the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals.
+
+The following implementation of a `TokenBuilder` will do the job for us. It creates two lexing modes, which are almost identical except for the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals.
+We will also need to make sure that the modes are switched based on the `TEMPLATE_LITERAL_START` and `TEMPLATE_LITERAL_END` terminals. We use `PUSH_MODE` and `POP_MODE` for this.
+
 ```ts
-import { DefaultTokenBuilder, Grammar, isTokenTypeArray, Keyword, TerminalRule } from "langium";
+import { DefaultTokenBuilder, isTokenTypeArray, GrammarAST } from "langium";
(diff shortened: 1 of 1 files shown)


hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+}
+```
+
+The `} ... {` block in this example won't be parsed as separate `}` and `{` tokens, but instead as a single `TEMPLATE_LITERAL_MIDDLE` token, resulting in a parser error due to the unexpected token.
+This doesn't make a lot of sense, since we aren't in the middle of a template literal at this point anyway.
+However, our lexer doesn't know yet that the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals are only allowed to show up within a `TemplateLiteral` rule.
+To rectify this, we will need to make use of lexer modes. They will give us the necessary context to know whether we're inside a template literal or outside of it.
+Depending on the current selected mode, we can lex different terminals. In our case, we want to exclude the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals.
+
+The following implementation of a `TokenBuilder` will do the job for us. It creates two lexing modes, which are almost identical except for the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals.
```
montymxb, line 80: There should be at least another sentence devoted to what the TokenBuilder is, and another one that describes what the 2 lexing modes are in this context.

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+}
+```
+
+With this change in place, the parser will work as expected. There is one last issue which we need to resolve in order to get everything working perfectly.
+When inspecting our AST, the `TemplateLiteral` object will contain strings with input artifacts in there; mainly `` ` ``, `{` and `}`.
+These aren't actually part of the semantic value of these strings, so we should get rid of them.
+We will need to create a custom `ValueConverter` and remove these artifacts:
+
+```ts
+import { CstNode, GrammarAST, DefaultValueConverter, ValueType, convertString } from 'langium';
```
montymxb, line 151:
```suggestion
import { CstNode, GrammarAST, DefaultValueConverter, ValueType } from 'langium';
```

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+
+export class CustomValueConverter extends DefaultValueConverter {
+
+    protected override runConverter(rule: GrammarAST.AbstractRule, input: string, cstNode: CstNode): ValueType {
+        if (rule.name.startsWith('TEMPLATE_LITERAL')) {
+            // 'convertString' simply removes the first and last character of the input
+            return convertString(input);
```
montymxb, line 158:
```suggestion
            // simply removes the first and last character of the input
            return input.slice(1,-1);
```
We can use slice here, in case we ever change the Langium API and `convertString` is changed or is removed. One less thing to maintain in the docs.

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+        } else {
+            return super.runConverter(rule, input, cstNode);
+        }
+    }
+}
+```
+
+Of course, let's not forget to bind all of these services:
```
montymxb, line 166:
```suggestion
Of course, let's not forget to bind all of these services in your **module.ts**:
```

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+
+```ts
+export const CustomModule = {
+    parser: {
+        TokenBuilder: () => new CustomTokenBuilder(),
+        ValueConverter: () => new CustomValueConverter()
+    },
+};
+```
```
montymxb, line 175: Would be good to add a concluding paragraph

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+```ts
+import { DefaultTokenBuilder, isTokenTypeArray, GrammarAST } from "langium";
+import { IMultiModeLexerDefinition, TokenType, TokenVocabulary } from "chevrotain";
+
+const REGULAR_MODE  = 'regular_mode';
+const TEMPLATE_MODE = 'template_mode';
+
+export class CustomTokenBuilder extends DefaultTokenBuilder {
+
+    override buildTokens(grammar: GrammarAST.Grammar, options?: { caseInsensitive?: boolean }): TokenVocabulary {
```
montymxb, line 92: I would break this out into a separate paragraph, explaining we need to first build up a multi-mode lexer definition that has various modes, which are pushed on by our special tokens.

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+                },
+                defaultMode: REGULAR_MODE
+            };
+            return multiModeLexerDef;
+        } else {
+            throw new Error('Invalid token vocabulary received from DefaultTokenBuilder!');
+        }
+    }
+
+    protected override buildKeywordToken(
```
montymxb, line 116: This would make a nice second part, indicating we need cleanup our `}` token so regular mode doesn't get messed up.

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+        
+        if (tokenType.name === '}') {
+            // The default } token will use [TEMPLATE_LITERAL_MIDDLE, TEMPLATE_LITERAL_END] as longer alts
+            // We need to delete the LONGER_ALT, they are not valid for the regular lexer mode
+            delete tokenType.LONGER_ALT;
+        }
+        return tokenType;
+    }
+
+    protected override buildTerminalToken(terminal: GrammarAST.TerminalRule): TokenType {
```
montymxb, line 131: Lastly, we can add this & explain how we're associating a push/pop action for start/end literals (which chevrotain needs).

montymxb (changes requested):
Cool, nice to see some stuff about multi-mode! However, this seems more about template literals than just multi-mode lexing. Not necessarily a problem, but I would recommend we change the title and make it clear that we're talking about implementing template literals through multi-mode lexing, as that appears to be the primary topic here. 

It might even make a better tutorial over a guide, as this is a targeted application; but that's less important than the first point.

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+title: "Multi-Mode Lexing"
+weight: 400
+---
+
+Many modern programming languages such as [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) or [C#](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) support template literals.
+They are a way to easily concatenate or interpolate string values while maintaining great code readability.
+This guide will show you how to support template literals in Langium.
+
+For this specific example, our template literal starts and ends using backticks `` ` ``  and are interupted by expressions that are wrapped in curly braces `{}`.
+So in our example, usage of template literals might look something like this:
```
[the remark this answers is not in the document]
montymxb: Can drop this one

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+
+```js
+println(`hello {name}!`);
+```
+
+Conceptually, template strings work by reading a start terminal which starts with `` ` `` and ends with `{`, 
+followed by an expression and then an end terminal which is effectively just the start terminal in reverse using `}` and `` ` ``.
+Since we don't want to restrict users to only a single expression in their template literals, we also need a "middle" terminal reading from `}` to `{`.
```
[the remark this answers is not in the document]
montymxb: Resolving as less relevant.

hugo/content/guides/multi-mode-lexing.md
```diff
@@ -0,0 +1,175 @@
...
+This doesn't make a lot of sense, since we aren't in the middle of a template literal at this point anyway.
+However, our lexer doesn't know yet that the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals are only allowed to show up within a `TemplateLiteral` rule.
+To rectify this, we will need to make use of lexer modes. They will give us the necessary context to know whether we're inside a template literal or outside of it.
+Depending on the current selected mode, we can lex different terminals. In our case, we want to exclude the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals.
+
+The following implementation of a `TokenBuilder` will do the job for us. It creates two lexing modes, which are almost identical except for the `TEMPLATE_LITERAL_MIDDLE` and `TEMPLATE_LITERAL_END` terminals.
+We will also need to make sure that the modes are switched based on the `TEMPLATE_LITERAL_START` and `TEMPLATE_LITERAL_END` terminals. We use `PUSH_MODE` and `POP_MODE` for this.
+
+```ts
```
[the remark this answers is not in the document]
montymxb: As a step up from this, I still feel we should split this up. But in the interest of moving this along after some time can we instead make an issue for a custom token builder guide separately?

hugo/content/guides/multi-mode-lexing.md
montymxb: See above.

hugo/content/guides/multi-mode-lexing.md
montymxb: See above.

hugo/content/guides/multi-mode-lexing.md
montymxb: See above.

hugo/content/guides/multi-mode-lexing.md
montymxb: Less important, will resolve in favor of grammar changes.

montymxb (commented):
Went back through and resolved a number of discussions to try and make the outstanding points clearer. Most of the remaining suggestions are specifically for grammar or clarity, but the rest should be good.

agacek:
The generated AstNode for TemplateLiteral looks like this:

```
export interface TemplateLiteral extends AstNode {
    ...
    content: Array<Expr> | Array<string>;
}
```

I would have expected to see `content: Array<Expr | string>`. Is this a bug or am I missing something?

msujew:
@agacek Thanks for the info, I've created https://github.com/eclipse-langium/langium/pull/1506 for this.

ym-han:
Just a quick piece of feedback: I skimmed through the tutorial in the PR, and noticed there wasn't actually much on *what*, conceptually speaking, a 'lexer mode' is. It may be worth adding a quick sentence explaining that when first introducing the concept -- presumably it's just state that the lexer uses to figure out what to do?

lars-reimann:
Really helpful guide, thanks. If the guide becomes primarily about template strings, it might be useful to briefly discuss how to get them working for languages with curly braces in an expression context (e.g. object literals of JS or dictionary literals in Python). Currently, code like `` `{ {"hello": "world"} }` `` would not be parsed correctly. Switching back to the regular mode when encountering an opening curly brace should resolve this:

```ts
    protected override buildKeywordToken(
        keyword: GrammarAST.Keyword,
        terminalTokens: TokenType[],
        caseInsensitive: boolean
    ): TokenType {
        let tokenType = super.buildKeywordToken(keyword, terminalTokens, caseInsensitive);
        
        if (tokenType.name === BLOCK_START) {
            // Enter regular mode (for object literals etc.)
            tokenType.PUSH_MODE = REGULAR_MODE;
        } else if (tokenType.name === '}') {
            // Return to previous mode
            tokenType.POP_MODE = true;

            // The default } token will use [TEMPLATE_LITERAL_MIDDLE, TEMPLATE_LITERAL_END] as longer alts
            // We need to delete the LONGER_ALT, they are not valid for the regular lexer mode
            delete tokenType.LONGER_ALT;
        }
        return tokenType;
```

PR #133, opened 2023-02-15
Sentence about glossary commented out.

jbicker:
Since the glossary doesn't exist the sentence about looking into it should not be shown


msujew (approved):
Thanks!

PR #134, opened 2023-02-27
Add buttons to homepage

msujew:
Adds a `Try it` and `Getting Started` button to the homepage. Also makes the header sticky (on desktop).


spoenemann (approved):
Looks good!

PR #140, opened 2023-03-17
Use prism for syntax highlighting

msujew:
Follow up on https://github.com/langium/langium-website/pull/125. This introduces a Prism based syntax highlighter that is specifically crafted for Langium.

Note that the statically baked in highlighting ([chroma based](https://github.com/alecthomas/chroma)) doesn't support extensions unless contributed to their repo, which is why I propose to use Prism in the first place.


danieldietrich (approved):
Hi Mark, thank you for this great addition, finally, *.langium code gets real syntax highlighting :) I performed the following checks during review:

* ✅ whitesmoke test: test page does still work as expected
* ✅ the markdown files do not contain code blocks highlighted with the `antlr` language
* ✅ the code blocks are highlighted consistently and the syntactic structure is well recognized by the highlighter (keywords, punctuation etc.)

I like the colors and the highlighting very well! Thx :)

PR #145, opened 2023-04-21
Minor fixes for the builtin library guide

felix-oq:
While following the guide for builtin libraries, I noticed a few places for potential improvements:

- Most importantly: add a type definition for `HelloWorldSharedServices`, because such a type is not generated by yeoman but required for the definition of `HelloWorldSharedModule`.
- Determine the file size via `Buffer.from(...).length`, so the actual byte size of the string is given.
- Add a call `super.loadAdditionalDocuments(folders, collector)`, for the case that the super class defines some additional behavior we might want to keep (though currently not the case).

Let me know what you think about these changes.

### DIFF
--- a/hugo/content/guides/builtin-library.md
+++ b/hugo/content/guides/builtin-library.md
@@ -53,6 +53,8 @@ As a last step, we have to bind our newly created workspace manager:
 
 ```ts
 // Add this to the `hello-world-module.ts` included in the yeoman generated project
+export type HelloWorldSharedServices = LangiumSharedServices;
+
 export const HelloWorldSharedModule: Module<HelloWorldSharedServices, DeepPartial<LangiumSharedServices>> = {
     workspace: {
         WorkspaceManager: (services) => new HelloWorldWorkspaceManager(services)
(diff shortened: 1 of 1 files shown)


hugo/content/guides/builtin-library.md
```diff
@@ -53,6 +54,8 @@ As a last step, we have to bind our newly created workspace manager:
 
 ```ts
 // Add this to the `hello-world-module.ts` included in the yeoman generated project
+export type HelloWorldSharedServices = LangiumSharedServices;
+
 export const HelloWorldSharedModule: Module<HelloWorldSharedServices, DeepPartial<LangiumSharedServices>> = {
```
msujew, line 59: Suggestion: The first and second type arguments shouldn't deviate here. In case the dev changes `HelloWorldSharedService`, the following is better:
```suggestion
export const HelloWorldSharedModule: Module<HelloWorldSharedServices, DeepPartial<HelloWorldSharedServices>> = {
```

msujew (commented):
Hey @felix-oq thank you for your contribution :)

hugo/content/guides/builtin-library.md
felix-oq: Good point, thanks!

msujew (approved):
This kind of fell under the radar. Looks good to me 👍

PR #146, opened 2023-04-27
replace support link with the new and more useful chat link

trusktr:
The support link was a little confusing: it starts with a Langium feature intro that has a URL that links back to the Langium site where we just came from. People clicking "support" will find the chat a lot more handy.


spoenemann:
The support link now leads to the newly designed https://www.typefox.io/language-engineering/ – I hope this page is more helpful for those seeking professional support.

Side note: The Langium project is now under vendor-neutral governance (Eclipse Foundation). This means that if other companies want to offer professional support for Langium and start contributing to the project, we can add them to the "Support" link as well. See for example how it's done for the Eclipse Theia project: https://theia-ide.org/support/

PR #157, opened 2023-06-12
Bump to monaco-editor-react 1.0.0

montymxb:
This updates our usage of the monaco-editor-react component to the latest 1.0.0 release. This required some adjustment to our existing implementation, but the end result should be easier to maintain down the road (and more flexible to modify as well).

There are still a few points that need to be resolved in the docs, and some other TODOs, before this is ready for review.


kaisalmen:
@montymxb do you want to close this in favor of #171 ?

montymxb:
Definitely, the other branch was already prepared and ahead of this one anyways.

PR #158, opened 2023-06-21
Update esbuild config for new versions

msujew:
With newer versions of esbuild, the build/watch API has changed quite a bit. This change updates our documentation to reflect that.

### DIFF
--- a/hugo/content/guides/code-bundling.md
+++ b/hugo/content/guides/code-bundling.md
@@ -39,8 +39,8 @@ if (watch) {
     await ctx.watch();
 } else {
     await ctx.rebuild();
+    ctx.dispose();
 }
-ctx.dispose();
 ```
 
 Store it in a module JavaScript file (`.mjs`) and create a corresponding script in your `package.json` file:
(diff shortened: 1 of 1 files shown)


hugo/content/guides/code-bundling.md
```diff
@@ -14,34 +14,44 @@ We generally recommend using [esbuild](https://esbuild.github.io/) to bundle Lan
...
-})
-    .then(() => console.log(success))
-    .catch(() => process.exit(1));
+});
+
+if (watch) {
+    await ctx.watch();
+} else {
+    await ctx.rebuild();
+    ctx.dispose();
```
spoenemann, line 42: BTW why is `dispose` only called in this case?

spoenemann (approved):
Thanks!

hugo/content/guides/code-bundling.md
msujew: The documentation actually says to do it for both cases, but for some reason it makes the `ctx.watch` call stop, and I have no idea why. Given that `watch` runs anyway until the process ends, there aren't any negative consequences of not calling `dispose`, so I'd guess it's fine that way.

PR #170, opened 2023-07-19
Handle Failed Editor Disposal

### LINKED ISSUE #163: Playground Bug when Identical terminal rules submitted.
While developing a JavaDoc Extractor to test Langium, It was found that while inserting 2 terminal rules with the same profile caused no error message to appear, and the Content to be 'Loading...' forever. 

The grammar is under development, hence the clear failure in contradictory statements, but should not cause the playground to crash in this way.

### Recording of the Bug
[Screencast from 2023-07-14 19-13-44.webm](https://github.com/langium/langium-website/assets/90779026/69bf821f-d860-4148-8e42-2e2271012ac2)

### Text of Grammar
Note that the terminal rules are identical barring the 'hidden'. 

```
grammar JavaDocExtractor

entry Model:
    (jdocs+=JDoc)*;
terminal JDoc: ('/**' -> '*/');

hidden terminal NotJDoc: ('/**' -> '*/');

```

### Text Content

```
/** foo 1 */
person John
person Jane

/* foo 2*/

Hello John!
Hello Jane!

/** foo 4*/
```

### Browser & OS

Google Chrome Version 114.0.5735.198 (Official Build) (64-bit)
OS Ubuntu 23.04
Linux Kernel 6.2.0-24-generic

montymxb:
Closes #163 , where the second program editor instance would crash due to not getting a correct grammar from the previous editor. In such a case there was no program editor to dispose of for future grammars, which would block the program editor from being setup again from that point forward. This just catches errors during the editor disposal process (such as trying to dispose a non-running editor), reports them, and unblocks subsequent setup of a new program editor.

Quick gif of the result below.
Jul-19-2023 17-05-46

### DIFF
--- a/hugo/content/playground/common.ts
+++ b/hugo/content/playground/common.ts
@@ -222,7 +222,12 @@ const PlaygroundActions: Actions = {
 
     if (editor) {
       content = editor.editor.getMainCode();
-      await editor.editor.dispose();
+      // attempt to dispose
+      await editor.editor.dispose().catch((e) => {
+        // report & discard this error
+        // can happen when a previous editor was not started correctly
+        console.error(e);
+      });
     }
 
     const { Grammar } = await createServicesForGrammar({ grammar: message.grammar });


msujew (approved):
Looks good to me 👍

PR #171, opened 2023-08-01
Upgrade Showcase/Playgrounds Monaco React/Wrapper Usage

montymxb:
This updates the showcase & playgrounds to use new versions of the `MonacoEditorLanguageClientWrapper` & `MonacoEditorReactComp` instances. The main change was the API interface for setting up the wrapper & the React component, which is composed of a more direct configuration object object, rather than several smaller configuration functions. This allowed us to have more control over the existing implementations, as well as to cut down on the configuration code for most cases.

The playgrounds required a rework to be compatible with the new approach, which exposed fewer low-level details than previously available. The new playgrounds approach has 2 wrappers like before, but the orchestration between the two underlying language servers is much simpler, and there's no longer a need to interact with the low-level message readers/writers. The current orchestration is setup like follows:
1. Langium LS receives a new Langium DSL
2. Langium LS parses & signals Langium LC on validating the new Langium DSL
3. Langium LC prepares a fresh worker using the new grammar
4. Langium LC regenerates the existing DSL Editor (2nd editor) w/ the new worker
5. DSL LS receives a new program, parses it, validates it
6. DSL LC is notified of a new program, and passes the resulting AST to the final window for rendering

I've tried to make the process as explicit as possible in the comments, but feel free to dig into things as needed to get more details. Overall, despite tearing down the 2nd DSL editor on getting a fresh grammar, the new approach is *very* fast. Application response times are about an order of magnitude quicker than they were before, and can probably be improved further if we really want to. An example of this can be seen below.

ex

A few checks that have been done so far:
- Mac
    - Safari
    - Chrome
    - Firefox

I haven't done any checks on Windows yet, so any insight here would be helpful to verify there's no bundle related issues.

I'm also open to suggestions, improvements, etc.!

TODO:
- ~README needs updating to follow new usage~
- ~Rebasing needed, and the new showcase examples need to be upgraded as well following this~


kaisalmen:
@montymxb works nicely on Windows with Chrome and Firefox

kaisalmen:
Observation: Where is greensock used? Can we get rid of those warnings or should I open an independent issue as this problem is not new or introduced here.
image

montymxb:
@kaisalmen not sure, but I don't believe it's new.

hugo/assets/scripts/mer.ts
```diff
@@ -0,0 +1,2 @@
+export * from "@typefox/monaco-editor-react";
```
Lotes, line 1: `mer`? Ah! `monaco-editor-react`?! Please do not use abbreviations, it requires readers to look it up or guess.

hugo/content/playground/data.ts
```diff
@@ -92,6 +92,7 @@ export const LangiumMonarchContent = {
   },
 };
 
+// hello world grammar
 export const LangiumInitialContent = `grammar HelloWorld
```
Lotes, line 96: Redundant information. Rename the variables instead of adding 2 comments.

hugo/content/playground/user-worker.ts
```diff
@@ -4,37 +4,77 @@
...
+        }
+        await startWithGrammar(event.data.grammar as string);
+    }
+});
+
+// disposable to remove previous build phase listeners
+let buildPhaseListener: Disposable;
+let newGrammarListener: Disposable;
+
+type DocumentChange = { uri: string, content: string, diagnostics: Diagnostic[] };
```
Lotes, line 24: `DocumentChange` & `DocumentChangeNotification`. Can they be moved into a common file? Saw them twice.

hugo/content/playground/utils.ts
```diff
@@ -15,27 +15,6 @@ export async function share(grammar: string, content: string): Promise<void> {
   await navigator.clipboard.writeText(url.toString());
 }
 
-export function throttle<T>(milliseconds: number, action: (input: T) => void) {
```
Lotes, line 18: Throttling is a way to minimize the rebuilds of the grammar. I used it to avoid consuming too much resources. Otherwise it would be after each keystroke triggered document build.

Lotes:
I like React function components more: https://www.robinwieruch.de/react-function-component/

But it would require to rewrite some stuff. The only, reeal benefit is shorter and more readable code.

hugo/content/playground/user-worker.ts
montymxb: Probably, I can look into that.

hugo/content/playground/utils.ts
montymxb: We still have this, it's been unrolled from the function to a couple of cases where `setTimeout` and `clearTimeout` are invoked directly (in common.ts). The current delays are 200ms as well, otherwise yeah we would have way too many requests.

Lotes: Great!

montymxb: I think while I was hacking away at things I pulled this out, but it might be nice to bring it back, and use the single function to replace the two usages, since it's effectively the same.

hugo/content/playground/data.ts
montymxb: New commit has this in

hugo/content/playground/user-worker.ts
montymxb: It's been deduped now. Also noticed that there was some old logic that was able to be pruned, slimming down both workers even more 😎 .

hugo/content/playground/utils.ts
montymxb: @Lotes I went ahead and brought back a version of `throttle` to manage the timeouts. Internally I keep information on previous timeouts, so there's callbacks to manage the state there anymore. Works nicely 🙂 .

kaisalmen:
@montymxb I checked both tutorial updates and you did a great job! 🎉

hugo/package.json
```diff
@@ -45,14 +47,15 @@
     "langium-minilogo": "^1.2.0",
     "langium-statemachine-dsl": "~1.2.0",
     "lz-string": "^1.4.4",
-    "monaco-editor-workers": "0.34.2",
-    "monaco-editor-wrapper": "1.4.1",
+    "monaco-editor-workers": "0.38.0",
```
kaisalmen, line 50: Please use `0.39.0`

package.json
```diff
@@ -20,7 +20,7 @@
     "tailwind"
   ],
   "volta": {
```
kaisalmen, line 22: node `18.17.0` is now available. It comes with npm `9.6.7`. Then we are aligned with current node LTS (for the next two weeks or so 😅

kaisalmen (commented):
Some minor things should be adjusted, otherwise this is great!

kaisalmen:
And, @Lotes already provides very useful suggestions. It is sometimes good to be the second reviewer. 😅

hugo/package.json
montymxb: Updated in the most recent commit.

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -93,7 +95,7 @@ For convenience, we're going to use two helper libraries from npm that wrap arou
 - [monaco-editor-wrapper](https://www.npmjs.com/package/monaco-editor-wrapper)
 - [monaco-editor-workers](https://www.npmjs.com/package/monaco-editor-workers)
 
-Both these packages should be installed as dependencies for your language. In particular, it's important that you're using version **1.4.0** or later of the monaco-editor-wrapper.
+Both these packages should be installed as dependencies for your language. In particular, this guide will assume that you're using version **2.0.0** or later of the monaco-editor-wrapper package, and version **0.34.2** of the monaco-editor-workers package.
```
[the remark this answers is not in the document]
montymxb: Updated in the most recent commit

package.json
montymxb: Updated to 18.17 & 9.6.7 respectively.

kaisalmen (approved):
Superb work! Let's go. 🚀

cdietrich:
hi, can i somewhere find a complete hello world example for the new code.
i tried to adapt what the yeoman generator produces
https://github.com/cdietrich/langium-inject-files-webworker/compare/cd_tests
but the ui seems totally broken
maybe you have any hints what i overlooked
Bildschirmfoto 2023-08-08 um 17 31 08

montymxb:
Hi @cdietrich . Yeah the new version has a different API than before, so setting it up as it was done before won't work as expected. To assist with this, we did update the [Langium and Monaco](https://langium.org/tutorials/langium_and_monaco/) docs to reflect setting up with version 2.1.1 of the monaco-editor-wrapper (which also happens to be targeted towards MiniLogo as well). If you haven't read that since it changed, I would double check those new docs, it might help.

As for the MiniLogo repo itself, it has not yet been updated to use version 2.1.1 of wrapper. We haven't gotten to that yet, but that will naturally follow.

montymxb:
I took a look at your implementation, I think the issue is that you're opting to use a classic editor config (Monarch), [but you're not passing a languageDef](https://github.com/cdietrich/langium-inject-files-webworker/compare/cd_tests#diff-5526eaf2942219477d776f994d9b2f1e1d60d8b7a6f4b5374736f7402571de62R74-R76) to utilize for that config. You're registering the id, but not the actual grammar itself, which is necessary for that to work.

For reference on how to set that up, [you can see how we add the languageDef to our user configs for the playgrounds & showcases](https://github.com/langium/langium-website/blob/main/hugo/assets/scripts/utils.ts#L177). You'll need a monarch grammar handy to pass in, which you can generate locally via Langium by tweaking your [langium-config.json](https://github.com/langium/langium/blob/main/examples/arithmetics/langium-config.json#L11) to output that in addition to a TextMate grammar when generating (although I think that's enabled by default now with the latest yeoman generator).

Give that a shot, and let me know if you're still having issues after that.

cdietrich:
@montymxb i had left the monarch out cause we use semantic tokens anyway.
using the monarch grammar does not really solve the problem
Bildschirmfoto 2023-08-09 um 12 22 41

there is still the white small text-area
and also the file outline is not next to the editor but below

montymxb:
You're missing the necessary styles for the editor. They used to be added in via `MonacoEditorLanguageClientWrapper.addMonacoStyles('monaco-editor-styles')`, but now they need to be imported from the package itself directly, or they come along implicitly when you're using the bundled version. For your case, that means you can add the following to your `index.html` alongside the existing stylesheet.
```html
<link href="/monaco-editor-wrapper/assets/style.css" rel="stylesheet"/>
```

The wrapper's assets are in your public root, so that should give you the styles you need, and your layout should be corrected. Checked this out locally real quick and it seems to have done the trick.

We missed adding this detail in the new docs, so we need to update that as well to ensure others don't run into the same issue.

cdietrich:
thx. this is working.
(had experimented with copy and use ./packages/examples/libs/monaco-editor-wrapper/assets/style.css before)
as it was working in the `verify*html`s in the example

cdietrich:
i also assume the yeoman generator will be adapted anyway to reflect it.

kaisalmen:
> i also assume the yeoman generator will be adapted anyway to reflect it.

@cdietrich I opened https://github.com/eclipse-langium/langium/issues/1163 for this

PR #181, opened 2023-09-01
Test PR for preview

netomi:


netomi:
@spoenemann some test PR to check if the preview mechanism works. However as this PR comes from a fork the action to add a PR comment fails as PRs from forks dont get write tokens. There are workarounds for that, but for security reasons its best not to hand out write tokens to PRs from forks as everyone can create them.

I guess you will only have PRs from the repo itself?

spoenemann:
> I guess you will only have PRs from the repo itself?

There will often be contributions from forks. But we could change the repo settings to require approval by maintainers before  running Actions?

netomi:
The github-deploy-pages action uses internally the sticky-pull-request-comment action which adds comments to the PR itself upon completion. This action needs write permissions for for pull-requests. For PR from forks, GitHub does not issue write tokens during workflow run for good reason. There is a workaround that we also utilize in our validation scripts that can be found here: https://github.com/adoptium/.eclipsefdn/blob/main/.github/workflows/validate.yml

The idea is to use pull_request_target instead.  You should read https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ before using that approach.

netomi:
Also, as you removed the netlify configuration from the repo, should the netlify app be uninstalled as it is not used anymore?

It had some advantages compared to deploying to gh-pages though imho.

spoenemann:
Yes we can remove the Netlify app. And yes, Netlify is more comfortable and offers very nice deploy previews, but we'd like to move the deployment to the Foundation. If you offer Netlify deployments as alternative to GH Pages, we can consider that as well.

netomi:
The netlify app has been removed from the organization.

I am not aware of what the EF offers in terms of deployment environments for previews, best to raise a ticket in the HelpDesk.

netomi:
btw the link to the issue in the deploy preview workflow from forks is here: https://github.com/rossjrw/pr-preview-action/issues/3

spoenemann:
No solution in sight so far. I propose to remove the previews (#208) until someone comes up with a properly working preview feature.

PR #182, opened 2023-09-01
Dispose of Wrapper on Failed Start

montymxb:
This addresses an issue with the recent playground changes, where a failed wrapper start was not disposing of the wrapper afterwards. This led to some strange behavior when working with grammars that would produce an invalid LS.

The solution here removes a previous workaround, and instead disposes of the wrapper in the appropriate place.

There is still a remaining issue with worker teardown, which does not succeed when the LC is stuck in a 'starting' state (see https://github.com/TypeFox/monaco-components/issues/41). We're able to address this by first terminating the worker manually, then disposing the wrapper.

Note that the wrapper dispose will fail when the LC is stuck, but the teardown up to that point is sufficient to clean up the rest of the artifacts (editor & associated elements).

### DIFF
--- a/hugo/content/playground/common.ts
+++ b/hugo/content/playground/common.ts
@@ -178,16 +178,6 @@ export async function setupPlayground(
     // get a fresh client
     dslClient = dslWrapper?.getLanguageClient();
     if (!dslClient) {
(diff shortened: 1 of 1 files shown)


kaisalmen (approved):
LGTM

PR #192, opened 2023-10-05
Update Langium + Monaco Tutorials

montymxb:
This updates the last couple of tutorials for *Langium + Monaco* & *Generation in the Web* to use more recent versions of Langium (2.0.2) and the [MEW](https://www.npmjs.com/package/monaco-editor-wrapper) (3.1.0). This follows the recent changes to langium-minilogo to version 2.0.0, which serves as the basis for a good chunk of these tutorials as well.

It would be great to get suggestions for wording and general flow improvements; as well as any typos that snuck in.

As a side note, I've noticed code-blocks are not overflowing as expected (instead of scrolling along the horizontal, they simply overflow past the right-hand boundary). This is not ideal, and appears to be a CSS issue, but is unrelated to the changes here.

### DIFF
--- a/hugo/content/tutorials/generation_in_the_web.md
+++ b/hugo/content/tutorials/generation_in_the_web.md
@@ -5,391 +5,172 @@ weight: 7
 
 {{< toc format=html >}}
 
-*Updated on Aug. 2nd, 2023 for usage with monaco-editor-wrapper 2.1.1 & above.*
+*Updated on Oct. 4th, 2023 for usage with monaco-editor-wrapper 3.1.0 & above.*
 
-In this tutorial we'll be talking about how to perform generation in the web by executing a custom LSP command. There are multiple ways to hook into Langium to utilize the generator, such as by directly exporting the generator API. However by using the LSP as is, we can save ourselves the effort of doing additional work. By using an LSP command, we can quickly and easily integrate new functionality into our existing Langium + Monaco integration.
+In this tutorial we'll be talking about how to perform generation in the web by listening for document builder notifications. There are multiple ways to hook into Langium to utilize the generator, such as by directly exporting the generator API. However, by listening to notifications from the document builder, we can do this with less code. This lets us quickly integrate new functionality into our existing Langium + Monaco integration, and focus more on what we would want to do with the generated output.
 
-We'll assume that you've already looked over most of the other tutorials at this point. It is particularly important that you have a language with working generation, and have a working instance of Langium + Monaco for your language (or another editor of your choice). In the case that you don't have a language to work with, you can follow along with [MiniLogo](https://github.com/langium/langium-minilogo), which is the example language used throughout these tutorials.
+*(This tutorial previously utilized custom LSP commands to achieve the same goal of generation. This is still a valid approach, but we've found setting up listening for notifications this way is much more straightforward. We've implemented this in our own example languages as well, and would recommend it going forward.)*
 
-Since we're working with MiniLogo, we already know that our generated output is in the form of drawing instructions that transform some drawing context. The generated output that we've implemented so far consists of a JSON array of commands, making it very easy to interpret. Now that we're working in a web-based context, this approach lends itself naturally towards manipulating an HTML5 canvas.
+We'll assume that you've already looked over most of the other tutorials at this point. It is particularly important that you have a language with working generation, and have a working instance of Langium + Monaco for your language (or another editor of your choice). In the case that you don't have a language to work with, you can follow along with [MiniLogo](https://github.com/langium/langium-minilogo), which is the example language used throughout many of these tutorials.
+
+Since we're working with MiniLogo here, we already know that our generated output is in the form of drawing instructions that transform some drawing context. The generated output that we've implemented so far consists of a JSON array of commands, making it very easy to interpret. Now that we're working in a web-based context, this approach lends itself naturally towards manipulating an HTML5 canvas.
 
 The parts that we still need to setup are:
 
-- exposing the generator via a custom LSP command
-- invoking this custom command and getting the result
-- adding a way to translate the generated result into drawing on an HTML5 canvas.
+- handle document validations, and generate notifications with our generator output
+- listen for these notifications in the client, and extract the generated output
+- interpret the generated output as drawing commands, and update the canvas
 
-## Overview of LSP Commands
+## Handling Document Validations
 
-Based on the work done in previous tutorials, we already have set up a working generator with MinLogo. If you haven't already set this up you can go back to the [tutorial on generation](/tutorials/generation) and give it a look over. Continuing off of the code written in that tutorial, we want to factor out our existing generator (removing any non-web compatible dependencies, like 'fs'), and invoke it via a custom LSP command handler.
+This is the first step we'll need, since without being able to generate notifications in the first place we would have nothing to listen to.
 
-If you're not familiar with the LSP (or custom commands), that's perfectly fine. The LSP is just a protocol that defines how our client & server communicate with each other, and this works even when they're both in the same application. In our case, the server will be Langium, and the client will be Monaco. This protocol also defines a way to [describe commands](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#command), and to [execute those custom commands from the client](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand). So, we'll be having Monaco execute a custom command, Langium handling that command, and then getting the results returned to Monaco.
-
-Lastly, Langium itself provides an easy way to register custom handlers for these commands. Handlers are registered for commands by name, and are invoked when that command is received. There are a number of reasons why this is a powerful approach:
-
-- Clients can send commands without any knowledge of Langium's internals
-- Langium can handle commands without any knowledge of the client sending them
-- Commands can be executed server-side with *full access* to Langium's capabilities
-
-This effectively allows Langium to integrate with just about any application that is capable of working with the LSP (and sending custom commands). It also does this without requiring any tight dependencies on Langium itself, keeping your existing application logic separate from your Langium logic.
-
-## Adding a Generator Endpoint for the Web
-
-We'll start by adding a new file **src/web/index.ts** that will act as the generator endpoint for the web. This directory was created in the previous tutorial about running Langium + Monaco in the web, and should already contain an express `app.ts` configuration.
-
-Our new file will contain a single exported function as our entry point, which will be used by our command handler. For MiniLogo we'll call this function `parseAndGenerate`. Much like the name suggests, this function takes a concrete MiniLogo program, parses it, and then generates output from the corresponding AST. This will share some logic that was used with the CLI before, so the code should be familiar if you've read the tutorial on [customizing the CLI](/tutorials/customizing_cli).
-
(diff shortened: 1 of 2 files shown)


hugo/content/tutorials/generation_in_the_web.md
```diff
@@ -5,391 +5,172 @@ weight: 7
...
 
-We'll assume that you've already looked over most of the other tutorials at this point. It is particularly important that you have a language with working generation, and have a working instance of Langium + Monaco for your language (or another editor of your choice). In the case that you don't have a language to work with, you can follow along with [MiniLogo](https://github.com/langium/langium-minilogo), which is the example language used throughout these tutorials.
+*(This tutorial previously utilized custom LSP commands to achieve the same goal of generation. This is still a valid approach, but we've found setting up listening for notifications this way is much more straightforward. We've implemented this in our own example languages as well, and would recommend it going forward.)*
```
kaisalmen, line 12: This is not is a good idea.

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -5,43 +5,35 @@ weight: 6
 
 {{< toc format=html >}}
 
-*Updated on Aug. 2nd, 2023 for usage with monaco-editor-wrapper 2.1.1 & above.*
+*Updated on Oct. 4th, 2023 for usage with monaco-editor-wrapper 3.1.0 & above, as well as Langium 2.0.2*
```
kaisalmen, line 8: We should use at least 3.2.x as the langium generator uses it and the service init was changed between 3.1.0 and 3.2.0. This could prevent some confusion.

kaisalmen: But not merged, yet. Sorry forgot. Maybe bring this in and make a small update afterwards is better.

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -52,68 +44,105 @@ const messageWriter = new BrowserMessageWriter(self);
...
 
-For convenience, we're going to use two helper libraries from npm that wrap around some of Monaco's core functionality.
+For convenience, we're going to use the Monaco Editor Wrapper (MER) to wrap around some of Monaco's core functionality, along with the Monaco Editor Workers package to assist. These packages are both maintained by TypeFox, and are designed to make it easier to use Monaco in a web-based context. We'll be using the following versions of these packages:
 
-- [monaco-editor-wrapper](https://www.npmjs.com/package/monaco-editor-wrapper)
-- [monaco-editor-workers](https://www.npmjs.com/package/monaco-editor-workers)
+- [Monaco Editor Wrapper](https://www.npmjs.com/package/monaco-editor-wrapper) version **3.1.0**
+- [monaco-editor-workers](https://www.npmjs.com/package/monaco-editor-workers) version **0.39.0**
 
 Both these packages should be installed as dependencies for your language. In particular, this guide will assume that you're using version **2.1.1** or later of the monaco-editor-wrapper package, and version **0.39.0** of the monaco-editor-workers package.
```
kaisalmen, line 92: **2.1.1** must be **3.1.0**

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -52,68 +44,105 @@ const messageWriter = new BrowserMessageWriter(self);
...
 
-The last script is there to provide a convenient way to invoke all the intermediate build steps in sequence. However you'll want to wait before running the `build:web` script, as we still need to add our **static** assets to make that work.
+Both scripts reference *mjs* files that need to be added as well into the scripts folder:
 
-If you went with another editor you would want to make sure that the assets required for that editor will also be copied into **public** folder as part of your output.
+**scripts/prepare-public.mjs**
+
+```js
+import * as esbuild from 'esbuild'
+import shell from 'shelljs'
```
kaisalmen, line 112: Note: In the langium generator I just used node functions, but shelljs just make life a lot easier.

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -169,233 +214,368 @@ h1 {
...
-import { MonacoEditorLanguageClientWrapper } from './monaco-editor-wrapper/index.js';
-import { buildWorkerDefinition } from "./monaco-editor-workers/index.js";
+```ts
+import { MonacoEditorLanguageClientWrapper, UserConfig } from "monaco-editor-wrapper/bundle";
+import { buildWorkerDefinition } from "monaco-editor-workers";
+import { addMonacoStyles } from 'monaco-editor-wrapper/styles';
 
-buildWorkerDefinition('./monaco-editor-workers/workers', new URL('', window.location.href).href, false);
+/**
+ * Prepare to setup the wrapper, building the worker def & setting up styles
```
kaisalmen, line 301: This comment should be updated. This is basically setting up monaco's own workers and necessary style for monaco-editor.

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -169,233 +214,368 @@ h1 {
...
-    // the later works better for us in this case
-    workerConfigOptions: lsWorker
-};
-```
+```ts
+type WorkerUrl = string;
 
-We can also pass more explicit config options to `workerConfigOptions` if we don't want to create the worker ourselves. For example:
+/**
+ * Generalized configuration used with 'getMonacoEditorReactConfig' to generate a working configuration for monaco-editor-react
```
kaisalmen, line 432: You suddenly mention "monaco-editor-react". I guess this is a copy-paste error.

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -169,233 +214,368 @@ h1 {
...
+    monarchGrammar: any;
+}
 
-```js
-{
-    url: new URL('./minilogo-server-worker.js', import.meta.url),
-    type: 'classic',
-    name: 'minilogo-language-server-worker'
+/**
+ * Generates a UserConfig for a given Langium example, which is then passed to the monaco-editor-react component
```
kaisalmen, line 443: You suddenly mention "monaco-editor-react". I guess this is a copy-paste error.

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -169,233 +214,368 @@ h1 {
...
+
+    // generate langium config
+    return {
+        htmlElement: config.htmlElement,
+        wrapperConfig: {
+            editorAppConfig: {
+                $type: 'classic',
+                languageId: id,
+                useDiffEditor: false,
+                automaticLayout: true,
```
kaisalmen, line 460: Btw, this is optional and `automaticLayout` is used if not specified.

hugo/content/tutorials/langium_and_monaco.md
```diff
@@ -169,233 +214,368 @@ h1 {
...
+                worker: config.worker as Worker,
+                name: `${id}-language-server-worker`
+            }
+        }
+    };
 }
 ```
 
-Either case works, but by creating the worker in advance, we give ourselves the ability to directly interact with the worker/LS independent of the wrapper itself, and to even pre-configure it before use. This can hugely beneficial, especially if we expect to customize our LS on the fly.
+This particular UserConfig will be for configuring a classic editor, rather than a VSCode API editor. This is because we're using a Monarch grammar, which is not supported by the VSCode API. However, if we wanted to use a TextMate grammar, we could use the VSCode API type instead.
```
kaisalmen, line 486: What about:
This particular UserConfig will be for configuring a classic editor, rather than a VSCode extension based editor. This is because we're using a Monarch grammar, which is not supported by the later. However, if we wanted to use a TextMate grammar, we could should use the extension based configuration instead.

hugo/content/tutorials/generation_in_the_web.md
```diff
@@ -5,391 +5,172 @@ weight: 7
...
-            <img width="125" src="https://langium.org/assets/langium_logo_w_nib.svg" alt="Langium">
-        </footer>
-        <!-- Monaco Configuration -->
-        <script type="module" src="setup.js"></script>
-    </body>
-</html>
-```
+```ts
+let running = false;
+let timeout: NodeJS.Timeout | null = null;
```
kaisalmen, line 118: `NodeJS.Timeout` on the client side?

kaisalmen (commented):
Great work! Thank you. Just some minor findings, comments and correction proposals.

hugo/content/tutorials/langium_and_monaco.md
kaisalmen: You can copy text from the updated README here:
https://github.com/TypeFox/monaco-components/pull/54

The text/comment above was one of the drivers to make this clearer.

kaisalmen:
@montymxb it would be best to align this with #196 (=update/align tutorial after merge)

hugo/content/tutorials/generation_in_the_web.md
montymxb: It's used as part of the throttle mechanic, so no requests are dispatched until we have X amount of time delay following (200ms I think is the current value).

hugo/content/tutorials/langium_and_monaco.md
montymxb: I've noticed this in some of the new examples, but I need to apply it here, thanks for pointing that out.

hugo/content/tutorials/langium_and_monaco.md
montymxb: Good catch! Need to update that.

hugo/content/tutorials/generation_in_the_web.md
montymxb: To not list the previous approach?

hugo/content/tutorials/langium_and_monaco.md
montymxb: At this point it may already be in, but we'll see, depends if I can update all usages accordingly before the next update.

hugo/content/tutorials/langium_and_monaco.md
montymxb: Good suggestion, I'll add this in.

hugo/content/tutorials/generation_in_the_web.md
kaisalmen: Sorry, I don't know why I wrote "not". I actually wanted to express that it is a good idea.

hugo/content/tutorials/langium_and_monaco.md
kaisalmen: If possible, I would make 3.3.0 the common baseline (minilogo?, this and generator-langium).

hugo/content/tutorials/generation_in_the_web.md
kaisalmen: I was just wondering if the code above is browser-compatible.

montymxb: Good point, it should be since the types are erased, but yeah that's the wrong type for a browser-based context. It should be a number instead (positive int to be specific).

hugo/content/tutorials/langium_and_monaco.md
montymxb: Given minilogo itself hasn't been updated yet to 3.3.0, this feels like another cycle (albeit smaller). We can do another cycle to update MiniLogo with the new wrapper after, and then the docs as well, which shouldn't be bad. It will also ensure we get the docs updated sooner rather than later (we haven't used the approach outlined here for some time now, so we should get that up to date).

kaisalmen (approved):
LGTM

PR #196, opened 2023-10-15
Update to latest monaco-editor-wrapper and monaco-editor-react

kaisalmen:
Updates the user config / app init code and any dependent code. The whole init and utility code regarding `monaco-editor-wrapper` and `@typefox/monaco-editor-react` and its bundle generation has been moved to its own package: **langium-website-core**

Hugo code only uses that bundle which prevents any complex javascript issues when generating the site.

All showcases and the playground work again!

This PR now uses the final version `3.3.0` of `monaco-editor-wrapper` and `2.3.0` of `@typefox/monaco-editor-react`


foundation/package.json
```diff
@@ -0,0 +1,67 @@
...
+        "dist/index"
+      ]
+    }
+  },
+  "files": [
+    "dist",
+    "bundle",
+    "src",
+    "LICENSE",
+    "README.md"
```
montymxb, line 36: Should add this README into the package

foundation/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "langium-website-foundation",
```
montymxb, line 2: Foundation sounds like an organization, core might be better here.

hugo/assets/scripts/arithmetics/arithmetics-tools.tsx
```diff
@@ -1,4 +1,4 @@
-import { monaco } from "monaco-editor-wrapper/.";
+import { monaco } from "langium-website-foundation/bundle";
```
montymxb, line 1: langium-website-core/bundle might be good here.

hugo/package.json
```diff
@@ -6,38 +6,34 @@
   "private": true,
   "scripts": {
     "clean:static": "shx rm -rf ./static/showcase ./static/playground ./static/libs",
-    "copy:libs:prepare": "shx mkdir -p ./static/libs/monaco-editor-workers/workers",
-    "copy:monaco-editor-wrapper": "npm run copy:libs:prepare && shx cp -fr ../node_modules/monaco-editor-wrapper/bundle ./static/libs/monaco-editor-wrapper",
+    "copy:libs:prepare": "shx mkdir -p ./static/libs/monaco-editor-workers/workers && shx mkdir -p ./static/libs/langium-website-foundation",
```
montymxb, line 9: We don't need to create a directory for the langium-website-foundation, nothing is copied there.

package.json
```diff
@@ -5,22 +5,23 @@
...
-    "build": "npm run clean && npm run build --workspace=tailwind --workspace=hugo",
-    "clean": "shx rm -rf public && npm run clean --workspace=tailwind"
+    "watch": "npm run build --workspace foundation && concurrently -c gray,blue -k -n hugo,tailwind \"npm run watch --workspace hugo\" \"npm run watch --workspace tailwind\"",
+    "watch:gitpod": "npm run build --workspace foundation && concurrently -c gray,blue -k -n hugo,tailwind \"npm run watch:gitpod --workspace hugo\" \"npm run watch --workspace tailwind\"",
+    "build": "npm run clean && npm run build --workspace foundation --workspace=tailwind --workspace=hugo",
+    "clean": "shx rm -rf public && npm run clean --workspace foundation --workspace=tailwind"
   },
   "devDependencies": {
-    "concurrently": "~7.6.0",
+    "concurrently": "~8.2.1",
```
montymxb, line 14: Noting that concurrently v8.X.X dropped node12 support in favor of node18, which works just fine for us here.

montymxb (changes requested):
- We should update the README to explain the purpose of the new package, and how it fits into the greater flow.
- Add a README to the new package itself, just to summarize it very briefly

Separately, I would consider **core** instead of **foundation** for the package name.

hugo/assets/scripts/arithmetics/arithmetics-tools.tsx
kaisalmen: Changed to `core` in next commit ⬇️

hugo/package.json
kaisalmen: Thanks for spotting that.

foundation/package.json
kaisalmen: Changed to `core` in next commit ⬇️

package.json
kaisalmen: Thanks.

kaisalmen:
All things are addressed. `foundation` is renamed to `core`. README is added and a launch config for Chrome is there now, too.

montymxb (approved):
:octocat: approved! Changes look good, and verified locally.

PR #198, opened 2023-10-17
Fix typo in getting-started grammar

VitGottwald:


hugo/content/docs/getting-started.md
```diff
@@ -75,7 +75,7 @@ Here's the grammar that parses the previous text snippet:
 grammar HelloWorld
 
 hidden terminal WS: /\s+/;
-terminal ID: /[_a-zA-Z][\w_]*/;
+terminal ID: /[_a-zA-Z][\w]*/;
```
spoenemann, line 78: I didn't know the `\w` class includes the underscore. I think this could be improved in our main repo, too. We have one instance of that terminal rule [in our Yeoman template](https://github.com/eclipse-langium/langium/blob/a4c7a58e9a79199d0f69cd3a9479c6ddfe99de21/packages/generator-langium/templates/core/src/language/language-id.langium#L13), and also in the Langium grammar language and our example languages. If you like, you're welcome to make the improvement in those places as well.

spoenemann (approved):
Thanks for the contribution!

PR #199, opened 2023-10-17
Make code blocks vertically scrollable

VitGottwald:
To make it possible to read the docs on a phone


spoenemann:
This commit was already included in #198.

PR #200, opened 2023-10-18
Minor DomainModel Showcase fixes

emilkrebs:
Scaling is now improved:

image

Nested Elements are now also rendered:

image


hugo/assets/scripts/domainmodel/domainmodel-tools.ts
```diff
@@ -22,6 +13,24 @@ export function getTreeNode(ast: AstNode): TreeNode {
...
+    // create a TreeNode from a PackageDeclaration
+    const getPackageDeclarationTreeNode = (p: PackageDeclaration): TreeNode => {
+        return {
+            ...p,
+            children: p.elements.map(e => getTreeNode(e))
+        }
+    }
+
+    // create a TreeNode from any DomainModelElement
+    const getTreeNode = (e: DomainModelElement): TreeNode => {
```
Lotes, line 25: Please rename the helper method, it has the same name as the parent function. This becomes confusing, because you do a recursive call and it is not clear which one is meant.

Otherwise looks good to me.

Lotes (approved):
I have no other helpful comments. For me this looks good.

Issue #185, opened 2023-09-06
Make PR previews work


spoenemann:
PR previews using `rossjrw/pr-preview-action` don't work yet because they are deployed to subpaths like `https://eclipse-langium.github.io/langium-website/pr-preview/pr-181/`.

In addition, there's the permissions issue mentioned in #181.

netomi:
The permissions problem can be solved.
The main problem is that the current deployment mechanism uses GitHub Actions, whereas the pr-preview-action commits to the gh-pages branch which is the legacy way to deploying.

The 2 mechanisms dont work together, as GH will not deploy from the gh-pages branch anymore, thus anything committed there will not be visible.

netomi:
It might be possible to uses the deploy-pages action: https://github.com/actions/deploy-pages

it has a beta feature called preview.

I will test it out in a fork.

netomi:
Ok so I managed now to come up with a setup to deploy previews of your site.

Take a look at this PR:

https://github.com/netomi/langium-website/pull/6

It has a link to the preview site available at https://netomi.github.io/langium-previews/pr-previews/pr-6/

What happens:

- when a PR is opened, a preview workflow run that build the page from this PR (needed  to add relativeURLs = true to the config.toml file)
- the build page is transferred to the langium-previews repo to a branch 'previews' in a subdir pr-previews/pr-X
- the langium-previews repo has GH pages actives by deploying from the previews branch
- so everytime something is pushed to the previews branch, the page gets deployed (this might take 1-2 mins)

similarly, if a PR is closed, the preview site will be removed from the langium-previews repo again.

Something that I still have to test is to support PRs from outside collaborators. I think this will be possible using the pull request target trigger, but need to check it.

Maybe this solution is something you are looking for? In the long term, hopefully GitHub supports deploying of previews but there is no information when this will be available, for the time being this might be a viable solution.

spoenemann:
Wow, cool -- thanks!

So this would require to create a new repo `langium-previews` in the GitHub org?

netomi:
yes, a separate repo is required to host the preview deployments.

spoenemann:
Ok the solution looks good. Please apply it to this repo.

Thank you for the great support! 🙏

msujew:
@netomi Thanks a lot for looking into this! This seems to work pretty well. See [here](https://eclipse-langium.github.io/langium-previews/pr-previews/pr-132/). We can fix the showcase related issues on our own :)

[...]

PR #208, opened 2023-12-12
Removed non-functioning PR previews

spoenemann:
Since the PR previews are broken (#185), I propose to remove that workflow until we find a proper solution.


spoenemann:
There seems to be a solution on the way: https://github.com/eclipse-langium/langium-website/issues/185#issuecomment-1852726337

PR #211, opened 2023-12-14
Unpin the pr-preview-action

netomi:
Need to unpin the action as it checks out itself using the provided tag


msujew:
@spoenemann Yeah, we probably need to merge it. In the GitHub Action, it still uses the old code for this PR.

spoenemann (approved):

PR #216, opened 2023-12-22
Visualized the Grammar AST 

emilkrebs:
Added a `3d-force-graph` to visualize the playground ast. 

preview

You can open the visualizer by clicking this button:

Screenshot

This doesn't just look good, it might also give users an understanding of the structure of DSLs and the grammar.


spoenemann (commented):
Thanks, the integration works well. However, I don't see how this can be useful when writing a grammar. Maybe we can discuss it at [one of the next dev meetings](https://github.com/eclipse-langium/langium/discussions/564?sort=new).

PR #220, opened 2024-03-20
Update showcases and playground to Langium 3.0.0

Lotes:
- it was required to update the dependencies first (Langium/SQL + Langium/Minilogo)
- I moved from Monarch to TextMate for the Playground syntax highlighting to have a unified approach
- migrated imports to access only Langium exports and using the `.js` suffix
- cleanup not required files


hugo/content/playground/common.ts
```diff
@@ -220,10 +220,10 @@ async function getFreshDSLWrapper(
     languageId,
     code,
     worker,
-    monarchGrammar: generateMonarch(Grammar, languageId)
+    textmateGrammar: generateTextMate(Grammar, {id: languageId, grammar: 'UserGrammar'})
   }), htmlElement).then(() => {
```
kaisalmen, line 224: We should use `await` and wrap everything in a big try, then you can just return the `wrapper` and `undefined` then there is no need for `as ...` at the end if I am not mistaken.

hugo/content/playground/preprocess.ts
```diff
@@ -104,7 +103,7 @@ export function preprocessAstNodeObject(
   const properties: PropertyNode[] = Object.keys(node)
     .filter((n) => !n.startsWith("$"))
     .map((n) => {
-      const valueOrValues = node[n] as
+      const valueOrValues = ((node as any)[n]) as
```
kaisalmen, line 106: Does it make sense to extract a type here?

hugo/content/playground/textmate-generator.ts
```diff
@@ -0,0 +1,255 @@
+/******************************************************************************
+ * Copyright 2021-2022 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+******************************************************************************/
+//ATTENTION!!! This file is basically copied from langium-cli!
```
kaisalmen, line 6: Should we export that in `langium-cli`? It is currently not available because it is not part of any export, correct?

hugo/package.json
```diff
@@ -38,19 +39,19 @@
     "clsx": "^1.2.1",
     "d3": "^7.8.5",
     "hugo-extended": "~0.89.4",
-    "langium": "^1.2.0",
-    "langium-arithmetics-dsl": "~1.2.1",
-    "langium-domainmodel-dsl": "~1.2.0",
-    "langium-minilogo": "^1.2.0",
-    "langium-sql": "^0.2.3",
-    "langium-statemachine-dsl": "~1.2.0",
+    "langium": "~3.0.0",
```
kaisalmen, line 42: We should uniformly define dependencies that has overlaps in the dependency tree, either relaxed `^` or more strict `~`

hugo/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
kaisalmen, line 1: Shouldn't this inherit from the base file?

core/package.json
```diff
@@ -46,7 +46,7 @@
     "@types/react-dom": "~18.2.13",
     "@types/vscode": "~1.83.0",
     "typescript": "~5.2.2",
-    "vite": "~4.4.11"
+    "vite": "^4.5.2"
```
kaisalmen, line 49: You can advance to vite 5. Should work just fine.

kaisalmen (commented):
I left a couple of comments.

hugo/content/playground/common.ts
kaisalmen: This will be changed with #222 as the function changed anyway.

core/package.json
kaisalmen: This will be changed with #222

hugo/package.json
Lotes: What do you mean? How should my action look like?

Lotes: Ah I think I see what you meant ^^...

hugo/tsconfig.json
Lotes: There is no base file... I guess you mean the `core` package... I could refactor out a base file instead...

hugo/package.json
Lotes: Done

hugo/content/playground/textmate-generator.ts
Lotes: [On its way](https://github.com/eclipse-langium/langium/pull/1434)...

montymxb (commented):
Changes look good, but just need to wait for the textmate grammar export to be available, which we can then consume.

Lotes:
@montymxb PR is ready to be approved. Last insertion adds `langium-cli/textmate` as source for the syntax highlighter.

msujew (approved):
Looks good to me, thank you 👍

kaisalmen:
@Lotes or @msujew can we merge this. I would like to get #222 ready for review (it is just a rebase now).

PR #221, opened 2024-03-25
Remove "bleeding edge" notice

spoenemann:
Declared types are now a stable feature.


montymxb (approved):
Cool, make sense to put this change in now.

PR #230, opened 2024-05-15
Playground: Customize linking errors

### LINKED ISSUE #226: Playground displays misleading linker errors
When using the Langium Playground, you're really only editing a grammar and seeing how it parses a sample input into an AST - the linking stage doesn't seem relevant. But linker errors are displayed, which as someone experimenting with Langium for the first time was very confusing until I realised what was happening.

I spent over an hour trying to understand why my grammar and content were producing errors, including diving into the XText docs. I nearly gave up on Langium as it just seemed broken, even in very simple scenarios. It turns out these errors just aren't resolvable in the Playground, because you're only working on the grammar and not doing anything related to resolution or scope.

The Playground is very likely to be used by people trying out Langium for the first time, and displaying these errors is very likely to lead to frustration and people giving up.

## Steps To Reproduce

1. Load the Playground at https://langium.org/playground/
2. Put the following into the Grammar pane:
```
grammar MyLang

entry Model:
    body+=Statement+;

Statement: 'const' assignment=ConstantAssignment ';';
ConstantAssignment: name=ID '=' value=Expression;
Expression: value=NUMBER | reference=[ConstantAssignment];

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;
terminal NUMBER returns number: /[0-9]+(\.[0-9]*)?/;
```
3. Put the following into the Content pane:
```
const n=5;
const y=n;
```

## The current behavior

image

## The expected behavior

No error should be displayed - the grammar is valid and has been successfully parsed into an AST. The reference hasn't been resolved, but we don't expect it to be - you can't write linker code in the playground, so many valid examples will result in errors.

Lotes:
Addresses and closes #226 

For the moment I am using warnings instead of errors for failed linking.

This PR is in draft mode, so I will change the implementation along the discussion inside of the issue.

### DIFF
--- a/hugo/content/playground/user-validator.ts
+++ b/hugo/content/playground/user-validator.ts
@@ -1,0 +1,28 @@
+import { AstNode, AstNodeDescription, AstUtils, DefaultDocumentValidator, DefaultLinker, DiagnosticInfo, DocumentState, DocumentValidator, LangiumDocument, LinkingError, LinkingErrorData, ReferenceInfo, ValidationOptions } from "langium";
+import { LangiumServices } from "langium/lsp";
+import { Diagnostic } from "vscode-languageserver";
+
+export class PlaygroundValidator extends DefaultDocumentValidator {
+    constructor(services: LangiumServices) {
+        super(services);
+    }
+    protected override processLinkingErrors(document: LangiumDocument, diagnostics: Diagnostic[], _options: ValidationOptions): void {
+        for (const reference of document.references) {
+            const linkingError = reference.error;
+            if (linkingError) {
+                const info: DiagnosticInfo<AstNode, string> = {
+                    node: linkingError.container,
+                    property: linkingError.property,
+                    index: linkingError.index,
+                    data: {
+                        code: DocumentValidator.LinkingError,
+                        containerType: linkingError.container.$type,
+                        property: linkingError.property,
+                        refText: linkingError.reference.$refText
+                    } satisfies LinkingErrorData
+                };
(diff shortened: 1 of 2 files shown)


Lotes:
Please have a look at my wording. Maybe you find a better way of expressing the warning. Otherwise I would be happy about an approval

hugo/content/playground/user-validator.ts
```diff
@@ -0,0 +1,28 @@
...
+                    property: linkingError.property,
+                    index: linkingError.index,
+                    data: {
+                        code: DocumentValidator.LinkingError,
+                        containerType: linkingError.container.$type,
+                        property: linkingError.property,
+                        refText: linkingError.reference.$refText
+                    } satisfies LinkingErrorData
+                };
+                diagnostics.push(this.toDiagnostic('warning', `Linking failed. Please inject a custom scope provider. This is a limitation of the playground. To overcome this issue, please study the learning section in the Langium documentation (original error: ${linkingError.message})`, info));
```
msujew, line 24: Suggestion: I believe it makes more sense to include the original error at the start of the diagnostic. Also asking the user to inject a custom scope provider is a bit weird in the context of the playground, as this isn't possible there.
```suggestion
                diagnostics.push(this.toDiagnostic('warning', ${linkingError.message}. In case you want to adjust the linking rules, please consult the learning section in the Langium documentation.`, info));
```

Lotes: Done, thank you.

Lotes: Bildschirmfoto 2024-05-23 um 13 04 25

msujew (approved):
LGTM 👍

PR #232, opened 2024-05-16
Specify class member scoping order

msujew:
Improves readability of the class member scoping invocation graph by making clear which answer belongs to which request.

### DIFF
--- a/hugo/content/docs/recipes/scoping/class-member.md
+++ b/hugo/content/docs/recipes/scoping/class-member.md
@@ -124,24 +124,24 @@ sequenceDiagram
     participant T as Type System
     participant S as Scope Provider
     participant L as Lexical Scope
-    R->>T: (1) Ask for type of expression<br>`element.sub.name`
-    T->>S: (2) Ask for `name` node
-    S->>T: (3) Ask for `sub` type
-    T->>S: (4) Ask for `sub` node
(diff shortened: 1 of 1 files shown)


hugo/content/docs/recipes/scoping/class-member.md
```diff
@@ -124,24 +124,24 @@ sequenceDiagram
...
-    S->>T: Give `name` node
-    T->>R: Give `string` type result
+    R->>T: (1) Ask for type of expression<br>`element.sub.name`
+    T->>S: (2) Ask for `name` node
+    S->>T: (3) Ask for `sub` type
+    T->>S: (4) Ask for `sub` node
+    S->>T: (5) Ask for `element` type
+    T->>S: (6) Ask for `element` node
+    S->>L: (7) Ask for `element` node
+    L->>S: (7) Give `element` node
```
Lotes, line 134: Nice idea with the numbering. Is it intended to use the wording "Ask/Give"? "Ask/Query" and "Return" would make it more direct or explicit. Just an idea.
If it is asynchronous it might be "send back" instead of "give" or "return".

msujew: I agree, thanks for the feedback!

Lotes (approved):
LGTM :) Thanks

PR #219, opened 2024-03-13
Guide how to enable keywords as identifiers #218

JohannesMeierSE:
This PR contributes a guide to the Langium documentation, how to enable keywords to be used as regular names/identifiers/values for properties, according to #218.

It roughly explains the background for the issues and shows to how overcome them along the "hello world" example.

I am looking forward to your feedback!

PR #233, opened 2024-05-22
New recipe: Case-insensitivity

Lotes:
Explains how to implement case-insensitivity in Langium.

Adds an extra folder: to be filled with other "Lexis topics". I think we can split #219 into 3-4 independent recipes.


hugo/content/docs/recipes/lexis/case-insensitive-languages.md
```diff
@@ -0,0 +1,50 @@
...
+---
+
+Case-insensitive languages are more suitable for beginners, because they don't have to worry about the case of identifiers and keywords.
+
+In Langium, there are basically two options you can choose from:
+
+* you can either make Langium case-insensitive by configuration
+* or you can include case-insensitivity only where you need it
+
+In both ways you need to take care of identifiers and cross-references.
```
msujew, line 13: Suggestion: Given that the second one only applies to terminals and not keywords, I wouldn't say they're equal alternatives:
```suggestion
In case you want to provide your users the same flexibility with your language, there are different levels of case-insensitivity in Langium:

* You can make Langium's parser completely case insensitive using the language configuration
* You can include case-insensitivity for specific terminal rules
* You can make cross references case insensitive

All of these options can be enabled independent of one another.
```

Lotes: I like especially the last sentence :D Thank you

hugo/content/docs/recipes/lexing/case-insensitive-languages.md
```diff
@@ -0,0 +1,54 @@
...
+    ],
+    ...
+}
+```
+
+## Case-insensitivity on demand
+
+If you want to include case-insensitivity only where you need it, you can use the `i` flag inside of your grammar's regular expressions
+
+```ts
```
msujew, line 39: FYI, we support `langium` as a syntax highlighting configuration on our website
````suggestion
```langium
````

msujew (approved):
Aside from the comment above, this looks good to me 👍

PR #235, opened 2024-05-23
Fix documentation linking

Lotes:
I still discovered some missing links since the refactoring of the docs.

I will continue on Monday to search for more gaps.


msujew (commented):
Question: Why keep `lexis` in the url for the lexing recipes?

Also, you missed one:
https://github.com/eclipse-langium/langium-website/blob/b2759fcaae2022da2fb5cfea22b8bebc39f71bf4/hugo/content/docs/reference/grammar-language.md?plain=1#L80

Otherwise, looks good to me 👍

Lotes:
I think it would be good to have a script doing this checking task.

Lotes:
@msujew I fixed the two findings of you, the remaining issues can be fixed in #237 .

msujew (approved):
LGTM 👍

PR #237, opened 2024-05-27
Add links checker and add aliases

Lotes:
This PR adds a checker for all links inside of Markdown files under `/hugo/content`.

I added aliases to the pages so that old links outside the webpage still work.

My checker has two levels:

* MISSING LINKS are links that point to no valid internal location
* ALIAS LINKs are links that point to an alias of an internal document, but not to its direct link

Both levels will lead to a failed build/check.

Comments/suggestions/improvements are welcome.


msujew (commented):
Looks pretty good.

msujew (approved):
LGTM, Thanks!

PR #238, opened 2024-05-30
Fix 404 page's styling

### LINKED ISSUE #236: Improve 404 Page and Dead Link Handling
After re-designing the layout of the documentation, we've ended up with a number of dead links. For example, the old getting-started link now redirects to a 404. This in itself is somewhat okay, but the 404 page is very broken looking, as can be seen below.
Screenshot 2024-05-24 at 10 06 40

The 404 page template needs to be corrected, and in cases where we know a link has moved, and it is appropriate to redirect to the new link location, we should do so.

Lotes:
Fixes #236

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Build
         run: |
           npm install
-          npm run build
+          BASE_URL=/pr-${{ github.event.number }} npm run preview
       - name: Upload artifact
         uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
         with:
--- a/hugo/config.toml
+++ b/hugo/config.toml
@@ -65,7 +65,7 @@ enableRobotsTXT = true
   # (Optional, default false) Set HTMl <base> to .Site.BaseURL if enabled. It might be required
   # if a subdirectory is used within Hugos BaseURL.
   # See https://developer.mozilla.org/de/docs/Web/HTML/Element/base.
-  geekdocOverwriteHTMLBase = true
+  #geekdocOverwriteHTMLBase = false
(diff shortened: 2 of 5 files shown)


Lotes:
This is still draft, because I need to test as a deployment ftm because on local mode it does not show up the custom 404. The problem so far is that the `others.html` partial includes paths to CSS that are not correct on subpages like "/docs/xxx"...

Lotes:
Bah, it seems I cannot test it really on a preview server. The 404 page is a result of the HTTP server, which redirects to `404.html` (GH-Preview-Pages does not). So what was missing? The base URL T_T...
I think when we release the changed flag, it will work. But I definitely cannot test it locally.

msujew (commented):
This seems to break the GitHub PR preview:
https://eclipse-langium.github.io/langium-previews/pr-previews/pr-238/docs/

Lotes:
> This seems to break the GitHub PR preview: https://eclipse-langium.github.io/langium-previews/pr-previews/pr-238/docs/

True. We need to set the correct base url before we are building. Otherwise it would be `https://eclipse-langium.github.io/langium-previews/pr-previews` instead of `https://eclipse-langium.github.io/langium-previews/pr-previews/pr-xxx`.

montymxb:
I've peeked at this too and it feels like hugo isn't responding to a myriad of ways to set a custom 404. The usual locations in layout, static, or even modifying the theme don't seem to have effect (notably the theme's own 404 isn't applied as expected). There can be a few reasons for this based on how we've customized the app, but we're also using an older version from about 3 years ago. I'm not sure how easily we could upgrade, but it might be worth checking if some newer version of hugo makes it easier to resolve this.

Lotes:
I am closing this PR. No idea how to check this, without breaking something.

PR #242, opened 2024-06-20
Removed outdated imprint, added Eclipse Foundation footer links

spoenemann:
TypeFox is no longer (solely) responsible for the website content. This change adapts the website to the new governance situation and uses the standard links of the Eclipse Foundation in the footer.


hugo/layouts/partials/langium-footer.html
```diff
@@ -1,20 +1,30 @@
...
+            Terms of Use
         </a>
     </div>
     <span class="text-lg hidden md:inline">|</span>
-    <div class="hover:underline px-4 block text-left md:text-center m-4" style="width:180px">
-        <a href="/imprint">
-            Imprint
+    <div class="px-4 block text-left md:text-center m-4" style="width:180px">
+        <a href="http://www.eclipse.org/legal/copyright.php" class="hover:underline">
+            Copyright Agent
```
danieldietrich, line 23: Why Copyright Agent and not just Copyright like above?
Never heard Agent before but I am sure it makes absolutely sense.
Thx

danieldietrich (approved):
Looks great, links are properly working, content is fine.
I did not test the Tailwind style on all devices, it works fine on my machine.
Thanks again, Miro!

PS: Out of curiosity, isn't the imprint not needed anymore because the Eclipse Fdn isn't a company?

hugo/layouts/partials/langium-footer.html
spoenemann: It's the same on other Eclipse Foundation websites.

spoenemann:
> PS: Out of curiosity, isn't the imprint not needed anymore because the Eclipse Fdn isn't a company?

I don't know – "imprint" could be a special requirement for Germany that doesn't apply elsewhere.

PR #246, opened 2024-07-18
Recipe: indentation-sensitive languages

aabounegm:
This is a guide on how to use the new `IndentationAwareTokenBuilder`, added to Langium in https://github.com/eclipse-langium/langium/pull/1578/ (published in `v3.2.0`).


hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,116 @@
...
+
+Some programming languages (such as Python, Haskell, and YAML) use indentation to denote nesting, as opposed to special non-whitespace tokens (such as `{` and `}` in C++/JavaScript).
+This can be difficult to express in the EBNF notation used for defining a language grammar in Langium, which is context-free.
+To achieve that, you can make use of synthetic tokens in the grammar which you would then redefine using Chevrotain in a custom token builder.
+
+Starting with Langium v3.2, such token builder (and an accompanying lexer) are provided for easy plugging into your language.
+
+## Configuring the token builder and lexer
+
+To be able to use the indendation tokens in your grammar, you first have to import and register the `IndentationAwareTokenBuilder` and `IndentationAwareLexer` services in your module as such:
```
Lotes, line 14: Could you link those two files, please. It would be interesting to see the implementation, even if it is already inside of Langium.

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,116 @@
+---
+title: Indentation-sensitive languages
+weight: 300
+---
+
+Some programming languages (such as Python, Haskell, and YAML) use indentation to denote nesting, as opposed to special non-whitespace tokens (such as `{` and `}` in C++/JavaScript).
+This can be difficult to express in the EBNF notation used for defining a language grammar in Langium, which is context-free.
+To achieve that, you can make use of synthetic tokens in the grammar which you would then redefine using Chevrotain in a custom token builder.
+
+Starting with Langium v3.2, such token builder (and an accompanying lexer) are provided for easy plugging into your language.
```
Lotes, line 10: Can you provide an explanation how this solution works. Other developers could benefit from those custom approach by imitating or adjusting the implementation.

Now it is just a blackbox. It is open source, yes, but it would be a good signal for others how to do things in Langium. Maybe a copy of the source code documentation is enough.

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,116 @@
...
+        INDENT elseBlock+=Statement+
+        DEDENT)?;
+
+Return: 'return' value=BOOLEAN;
+
+terminal BOOLEAN returns boolean: /true|false/;
+terminal INDENT: 'synthetic:indent';
+terminal DEDENT: 'synthetic:dedent';
+hidden terminal WS: /[\t ]+/;
+hidden terminal NL: /[\r\n]+/;
```
Lotes, line 63: Why do you have split WS and NL? Is this important in order to use this approach?

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+The important terminals here are `INDENT`, `DEDENT`, and `WS`.
+`INDENT` and `DEDENT` are used to delimit a nested block, similar to `{` and `}` (respectively) in C-like languages.
+Note that `INDENT` indicates an **increase** in indentation, not just the existence of leading whitespace, which is why in the example above we used it only at the beginning of the block, not before every `Statement`.
+
+The content you choose for these 3 terminals doesn't matter since it will overridden by `IndentationAwareTokenBuilder` anyway. However, you might still want to choose tokens that don't overlap with other terminals for easier use in the playground.
```
Lotes, line 70: As already stated: Developers could benefit from a summary how this approach is implemented. You release small pieces of information in this direction. But it would be really nice to have the whole strategy on the plate. So that it becomes more easy to adapt.
When the code is documented well, then a link to the code section might be sufficient... not sure.

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Playground compatibility
```
Lotes, line 72: Mh not sure if we need this. It should be clear for users that the playground is limited.

Lotes (commented):
Really nice recipe. I have some comments about the insights to the implementation. I think we should add links or detailed information how this token builder/lexer works.

aabounegm:
Thanks for your comments! I will address them as soon as a decision is reached about https://github.com/eclipse-langium/langium/pull/1608 to edit the last section as well

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
aabounegm: Yes, it is important because having a single `\s+` terminal will cause it to eat up the indentation along with the `\n` in the same token. Yes, it will have a lower priority than `INDENT`/`DEDENT`, but these 2 will not match at the end of the line, while `\s+` will.

aabounegm:
Thanks for waiting so long, all comments should be addressed now.
Note that the last commit adds a subsection that depends on https://github.com/eclipse-langium/langium/pull/1647 getting merged first

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,135 @@
...
+With the default configuration and the grammar above, for the following code sample:
+```
+if true:
+    return false
+else:
+    if true:
+        return true
+```
+
+the lexer will output the following sequence of tokens: `if`, `BOOLEAN`, `INDENT`, `return`, `BOOLEAN`, `DEDENT`, `else`, `INDENT`, `if`, `BOOLEAN`, `INDENT`, `return`, `BOOLEAN`, `DEDENT`, `DEDENT`.
```
Lotes, line 135: `The` with capital T

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
Lotes: You provided some links to the source, which is good.

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
Lotes: I think it is easier to follow. Interested persons can lookup the source directly.

Lotes (commented):
Just some minor things and one big question. The most of my thoughts were already resolved. Thanks.

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,135 @@
+---
```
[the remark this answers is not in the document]
aabounegm: That's a good question!
I thought a lot about the best approach here, and in the end decided not to discriminate between them, which is the simpler way. Alternatives included allowing only one or the other through a config parameter, or treating a tab as `n` spaces (again, for a configurable `n`). I thought these 2 alternatives were a bit too strict (though that's how Python behaves, for example, by prohibiting mixing them), and I thought that ideally I could issue a warning, but I couldn't find a way to accept a token and still issue a warning/error.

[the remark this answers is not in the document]
aabounegm: Actually, now that I think about it, I could add some payload to the returned token and then in the lexer check for the payload and add to the errors array, but then there would still be no way of making it a warning rather than an error. Perhaps `LexerResult` should be augmented to allow warnings?

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
aabounegm: It was intended as a continuation of the sentence before it, only interrupted by the code snippet. Not sure if it makes sense or if it counts as a separate sentence 🤔

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,135 @@
...
+```ts
+import { MyLanguageTerminalNames, MyLanguageKeywordNames } from './generated/ast.js';
+import { IndentationAwareTokenBuilder, IndentationAwareLexer } from 'langium';
+
+// ...
+export const HelloWorldModule: Module<HelloWorldServices, PartialLangiumServices & HelloWorldAddedServices> = {
+    parser: {
+        TokenBuilder: () => new IndentationAwareTokenBuilder<MyLanguageTerminalNames, MyLanguageKeywordNames>({
+            ignoreIndentationDelimiters: [
+                ['L_BRAC', 'R_BARC'], // <-- This typo will now cause a TypeScript error
```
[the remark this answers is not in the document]
aabounegm: Thank you 😁

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
```diff
@@ -0,0 +1,135 @@
...
+```python
+x = [
+    1,
+    2
+]
+```
+any indentation between `[` and `]` should be ignored.
+
+To achieve similar behavior with the `IndentationAwareTokenBuilder`, the `ignoreIndentationDelimiters` option can be used.
+It accepts is a list of pairs of token names (terminal or keyword) and turns off indentation token detection between each pair.
```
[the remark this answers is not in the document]
aabounegm: Right, not sure how it got there

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
aabounegm: Ok, removed this section

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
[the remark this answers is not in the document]
Lotes: I think resolution of my question would block this recipe. I mean we can still change something afterwards. Extending the LexerResult sounds too much for this change.

Another question could be: How to write an indention-aware formatter? Is it even applicable or doable? How is it done for Python?
We do not have to answer this now. I was just interested about some consequences or follow-up tasks.

Lotes (commented):
I have nothing more to add right now. Let's wait for a second opinion on your recipe :-) .

hugo/content/docs/recipes/lexing/indentation-sensitive-languages.md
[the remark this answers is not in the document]
aabounegm: I do not yet have experience writing formatters in Langium, but I don't see why it would be difficult to do. Generally, there are 2 approaches: formatting and pretty printing. One way to implement a formatter is to search for some (anti-)patterns in the code and issue `TextEdit`s just for them. Pretty printers normally use the AST/CST (or some other intermediate representation) and transform them back into code, regardless of how it initially looked like before parsing. (or at least that's how I understand the difference between them)
Both approaches seem possible with the indentation-aware tokens, though the second one (pretty printer) is probably easier to implement, assuming we want the formatter to ensure consistent indentation characters and sizes.

For Python, one of the most popular formatter is `black`, and it uses the pretty printing approach. Not sure how other formatters handle inconsistent indentation, tbh.

aabounegm:
@Lotes Langium v3.2 has already been published with the indentation-aware token builder, and questions about its usage already started coming in (https://github.com/eclipse-langium/langium/discussions/1696). Are we waiting for another review or can this recipe be merged?

aabounegm:
Sure! Replaced 👍

PR #249, opened 2024-08-22
Try to fix 404

Lotes:


hugo/package.json
```diff
@@ -18,7 +18,7 @@
     "build:worker/sql": "esbuild ./assets/scripts/sql/language-server.ts --bundle --format=iife --outfile=./static/showcase/libs/worker/sqlServerWorker.js",
     "build:worker/minilogo": "esbuild ../node_modules/langium-minilogo/out/language-server/main-browser.js --bundle --format=iife --outfile=./static/showcase/libs/worker/minilogoServerWorker.js",
     "build:static": "npm run clean:static && npm run build:worker/statemachine && npm run build:worker/domainmodel && npm run build:worker/sql && npm run build:worker/minilogo && npm run build:worker/arithmetics && npm run build:worker/playground-common && npm run build:worker/playground-langium && npm run build:worker/playground-user && npm run copy:monaco-workers",
-    "build": "npm run build:static && cross-env NODE_ENV=production  hugo        --config ./config.toml    -b /              -d ../public --gc --minify    ",
+    "build": "./scripts/build.sh production /",
```
msujew, line 21: This won't work on Windows :)

Lotes:
I tried to make the paths relative, but Hugo seems not to except some little amount of links and makes them relative to the root instead of the current path. That makes the link checker pointless.

aabounegm:
I think one issue could be that the base URL doesn't end in a trailing slash. I didn't run Hugo myself, but in the following HTML:
```html
<head>
  <base href="/pr-249">
</head>
<body>
  <a href="./page">Page</a>
  <a href="page">equivalent link</a>
</body>
```
the links will actually point to `/page`. To have them point to `/pr-249/page`, the base URL needs to be `/pr-249/` instead.

The other issue is that with `<base>` set, all relative URLs in the website become relative to that base. For example, `./features` will be `/pr-249/features`, not `/pr-249/docs/features` as intended.

Lastly, I think the build script is not necessary because Hugo can be configured through environment variables, according to [the docs](https://gohugo.io/getting-started/configuration/#configure-with-environment-variables). I believe it should suffice to set the env var `HUGO_BASEURL` instead.

PR #253, opened 2024-09-06
Replace prism with shiki

msujew:
Closes https://github.com/eclipse-langium/langium-website/issues/244

Allows us to more easily highlight our custom code (like the `minilogo` language and other in our documentation), using textmate instead of prism. Highlighting our custom code can be done in a separate PR, this only highlights the already existing code.


scripts/shikify.ts
```diff
@@ -0,0 +1,68 @@
...
+for (const file of htmlFiles) {
+    // Grab the file, and parse it into a DOM
+    const filePath = path.join(mainPath, file);
+    console.log(`Processing ${file}`);
+    const content = await fs.readFile(filePath, { encoding: "utf-8" });
+    const dom = parse(content);
+
+    // This isn't a particularly smart query implementation,
+    // so lets take the simple route and just grab all of the pre tags
+    const codeBlocks = dom.querySelectorAll("pre")
```
Lotes, line 31: Very pragmatic :D... but what about a more official way? Are there no `shiki` plugins or addons for Hugo?

I googled it and discovered your source for this code:
https://blog.puzzmo.com/posts/2024/06/23/shiki-hugo/
We should add a link in case of any improvements in their comments (Maybe we have some trouble with this code, then it is good to look up this link first).

WDYT?

package.json
```diff
@@ -4,12 +4,14 @@
   "author": "TypeFox",
   "license": "MIT",
   "private": true,
+  "type": "module",
```
Lotes, line 7: This destroys the link checking build step :-/...

PR #257, opened 2024-10-08
Link the API docs

### LINKED ISSUE #254: Add a page with API documentation
It would help a lot with exploring Langium and the functionalities it provides if there was a page in the documentation with the whole API exported from Langium. This is easy to do using the JSDoc comments with [TypeDoc](https://typedoc.org/) (I'm already preparing a PR), but some organizational questions need to be answered first.
Since the API documentation is generated from the source code, it should be published on new Langium releases, and not independently in the website repo. However, it should still be linked from the website, leading to the following questions:
1. Should the API docs live on https://langium.org/api or a separate subdomain like https://api.langium.org?
2. If the former, how will it be integrated with the website? Using an iframe or should it be built as part of the website's build process?
3. Do we care about keeping docs for old Langium versions, or just the latest?

aabounegm:
Closes #254

### DIFF
--- a/hugo/content/api/_index.md
+++ b/hugo/content/api/_index.md
@@ -1,0 +1,5 @@
+---
+title: "API"
+weight: 300
+---
+<meta http-equiv="refresh" content="0; url=https://eclipse-langium.github.io/langium/"/>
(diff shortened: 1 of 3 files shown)


msujew (approved):
Thanks, looks good to me 👍

PR #269, opened 2024-11-20
Fix import on keywords-as-identifiers recipe

leonardtudorache:
`langium` has no exported member _AbstractSemanticTokenProvider_ or _SemanticTokenAcceptor_. These should be imported from `langium/lsp`.


msujew (commented):
Hey @leonardtudorache, thank you for the contribution!

Due to legal reasons, contributors to Langium (or rather all Eclipse Foundation projects) must sign the [Eclipse Contribution Agreement](https://www.eclipse.org/legal/eca/) with the same email that was used to create the commit before we're able to merge the changes (no matter how small).

Can you do that? Otherwise I will just create a PR myself with the necessary changes and close this one.

msujew (approved):
Looks good to me! We wrote this guide before we split the LSP stuff from the main import in v3.

PR #277, opened 2025-04-08
formatting.md: Fix imports

borisbrodski:
The import statement in the example code was not working.

I changed it from

```
import { AbstractFormatter, AstNode, Formatting } from 'langium';
```

to

```
import { AstNode } from 'langium';
import { AbstractFormatter, Formatting } from 'langium/lsp';
```


msujew (approved):
Thanks!

PR #287, opened 2025-11-04
Add documentation for profiling

Lotes:
Closes #284 

There is one TODO, I cannot resolve: What do we measure in the validation phase?


msujew (commented):
Looks quite good. See the comment below for the open TODO item.

PR #289, opened 2026-06-30
Point to New & Improved MLC Guides

montymxb:
Clears out a pair of older Langium (Minilogo-based) guides that revolved around Langium + Monaco, and leaves links to the new content along with a brief explanation, mostly to preserve all existing incoming links.

These ones have been dated for a bit now, and we wanted to freshen them up. The new ones are located over at the [mlc's Langium guides section](https://github.com/TypeFox/monaco-languageclient/). Those focus on the same topics these two tutorials did, but with more emphasis on the mlc side of things.


msujew (approved):
Thanks!

PR #290, opened 2026-07-14
Fix domain-model example

msujew:
As of version 4 of Langium, each file (by default) needs the correct file extension. This change adjusts the generated URI to match that.


emilkrebs (approved):
LGTM! Seems like the failing action is not related to this change anyway so probably ready to merge

